PR #2, opened 2022-08-31
Adding automation and tests

Trompettesib:


huntears:
You may want to change the compiler for the CI from gcc to clang

Trompettesib:
How can I do to start another time this action ?

Trompettesib:
If this is possible, I want to remove the last two commit from this PR because they are only for GTests things and not about the "pure" CI. I would like to do another PR only for this two commits (if this is relevent of course)
I don't know how to do that ^^ this is why I write it there ^^

huntears (dismissed):
LGTM

PR #5, opened 2022-09-12
Create config.yml + files setup

NekoSelena:
Don't know if something is missing


huntears:
The config file looks good, everything is there and well sorted

PR #10, opened 2022-09-17
Use std::min instead of a ternary in Client::_sendData

### LINKED ISSUE #9: Use std::min instead of a ternary in Client::_sendData
https://github.com/CubicMC/cubic-server/blob/784091539cda8b704b2c1b66aa914a6011864fe4/cubic-server/Client.cpp#L81

huntears:
Fixes #9


Trompettesib (approved):
This is much better !

PR #15, opened 2022-09-24
Add Configuration - Handling the config.yml

NekoSelena:


NekoSelena:
The include is not found, while it works fine on my computer
I tried changing the "" by <>, but that's not the issue

Trompettesib:
@huntears We need help ^^

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
+#include "ConfigHandler.hpp"
+#include <iostream>
+#include <fstream>
+#include "../yaml-cpp/include/yaml-cpp/yaml.h"
```
Trompettesib, line 4: Include cringe but I don't know how to do it better

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+Configuration::ConfigHandler::ConfigHandler(std::string filePath)
+{
+    this->getConfigFile(filePath);
+}
+
+Configuration::ConfigHandler::~ConfigHandler(){}
```
Trompettesib, line 11: You can remove it if it's empty/default

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    const std::string ip = config["network"]["ip"].as<std::string>();
+    const int port = config["network"]["port"].as<int>();
+    const int max_players = config["general"]["max_players"].as<int>();
+    const std::string motd = config["general"]["motd"].as<std::string>();
+
+    this->_configFile.insert(std::pair<std::string, std::string>("ip", ip));
+    this->_configFile.insert(std::pair<std::string, std::string>("port", std::to_string(port)));
+    this->_configFile.insert(std::pair<std::string, std::string>("max_players", std::to_string(max_players)));
+    this->_configFile.insert(std::pair<std::string, std::string>("motd", motd));
```
Trompettesib, line 25: Maybe you can use [] operator.
I don't know if it's more opti

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+std::string Configuration::ConfigHandler::getIP(void)
+{
+    std::string value = this->_configFile.find("ip")->second;
```
Trompettesib, line 30: Maybe you can use [] operator.
I don't know if it's more opti

configuration/ConfigHandler.hpp
```diff
@@ -0,0 +1,31 @@
...
+        ConfigHandler(std::string);
+        ~ConfigHandler();
+
+        std::string getIP(void);
+        std::string getMotd(void);
+        std::string getPort(void);
+        std::string getMaxPlayers(void);
+
+    private:
+        std::map <std::string, std::string> _configFile;
```
Trompettesib, line 26: Maybe you can use an unordered_map instead of a map

configuration/ConfigHandler.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @brief Handles the server configuration
+     * by parsing the config.yml file
+     */
+    class ConfigHandler
+    {
+    public:
+        ConfigHandler(std::string);
+        ~ConfigHandler();
```
Trompettesib, line 18: You can remove it if it's a default/empty destructor

Trompettesib (commented):
LGTM

configuration/ConfigHandler.hpp
huntears: This would indeed get way better performance but it is only a config reader, so it's not that important

CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ endif()
 add_subdirectory (protocol)
 add_subdirectory (cubic-server)
 add_subdirectory (math)
+add_subdirectory (yaml-cpp)
```
huntears, line 39: Including a whole library directly into our project is really not the proper way to do it, either have CMake download it and set it up or have it be here as a git-submodule (A CMake dependency would be better).

config.yml
```diff
@@ -0,0 +1,7 @@
+network:
```
huntears, line 1: This file should not be at the of the project, but in a sub-directory.

configuration/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Create a library called "Hello" which includes the source file "hello.cxx".
+# The extension is already found. Any number of sources could be listed here.
+add_library (ConfigHandler ConfigHandler.cpp)
+set(CMAKE_CXX_FLAGS "-lyaml-cpp")
```
huntears, line 4: This is compiler dependant, please use `target_link_libraries`

configuration/ConfigHandler.cpp
huntears, line 4: CMake allows to setup proper include paths in the build system, this avoids doing relative import paths

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+    this->getConfigFile(filePath);
+}
+
+Configuration::ConfigHandler::~ConfigHandler(){}
+
+void Configuration::ConfigHandler::getConfigFile(std::string configFile)
+{
+    YAML::Node config = YAML::LoadFile(configFile);
+
+    const std::string ip = config["network"]["ip"].as<std::string>();
```
huntears, line 17: There is no error handling whatsover, please check the return values of the config access as the fields could not exist, as right now this would just trigger a segmentation fault.

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+    const int port = config["network"]["port"].as<int>();
+    const int max_players = config["general"]["max_players"].as<int>();
+    const std::string motd = config["general"]["motd"].as<std::string>();
+
+    this->_configFile.insert(std::pair<std::string, std::string>("ip", ip));
```
huntears, line 22: Please simplify this by using initializer list instead of creating a whole `std::pair` like this : 
```cpp
this->_configFile.insert({"ip", ip});
```

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "../yaml-cpp/include/yaml-cpp/yaml.h"
+
+Configuration::ConfigHandler::ConfigHandler(std::string filePath)
+{
+    this->getConfigFile(filePath);
+}
+
+Configuration::ConfigHandler::~ConfigHandler(){}
+
+void Configuration::ConfigHandler::getConfigFile(std::string configFile)
```
huntears, line 13: Please make each entry getter more generic, currently if we want to add other fields we have to add multiples lines to this one function, which will make the code unmaintainable.

configuration/ConfigHandler.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+std::string Configuration::ConfigHandler::getIP(void)
+{
+    std::string value = this->_configFile.find("ip")->second;
+    return value;
+}
+
+std::string Configuration::ConfigHandler::getMotd(void)
+{
+    std::string value = this->_configFile.find("motd")->second;
```
huntears, line 36: Please check the return value of find, if the key doesn't exist this will segfault.
(Same with the other getters)

configuration/ConfigHandler.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+        std::string getIP(void);
+        std::string getMotd(void);
+        std::string getPort(void);
```
huntears, line 22: Why is the port stored as an `std::string` and not an `uint16_t` ?

configuration/ConfigHandler.hpp
```diff
@@ -0,0 +1,31 @@
...
+        std::string getMaxPlayers(void);
```
huntears, line 23: Same there with the max player count, why is it an `std::string` and not an int ?

configuration/ConfigHandler.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    private:
+        std::map <std::string, std::string> _configFile;
+        void getConfigFile(std::string);
```
huntears, line 27: Use constant references when possible

huntears (changes requested):
A lot of things to change overall but we're getting somewhere :)

configuration/ConfigHandler.hpp
NekoSelena: Because I'm very bad at cpp, so everything is a std::string to make it easier for me

huntears: Well now is the time to try and make better cpp :)

configuration/CMakeLists.txt
NekoSelena: Do I use it like this ?
`target_link_libraries(yaml-cpp)`

NekoSelena: 'cause I don't quite understand the cmake doc about this

Trompettesib (approved):
GG WP
LGTM

huntears (approved):
LGTM

PR #16, opened 2022-09-24
Logs

Trompettesib:
Add logging class


logging/Logger_unittest.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        testing::internal::CaptureStdout();
+        logger.log(LogLevel::DEBUG, "not_a_test");
+        output = testing::internal::GetCapturedStdout();
+        std::getline(file, line);
+
+        EXPECT_EQ(line.find("not_a_test"), std::string::npos);
+        EXPECT_NE(output.find("not_a_test"), std::string::npos);
+
+        // file.seekg(0, std::ios::beg);
```
Trompettesib, line 104: Why isn't this working ?

logging/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# The extension is already found. Any number of sources could be listed here.
+add_library (TimeFormatter TimeFormatter.cpp)
+add_library (FileAndFolderHandler FileAndFolderHandler.cpp)
+add_library (Logging Logger.cpp)
+
+# Make sure the compiler can find include files for our Hello library
+# when other libraries or executables link to Hello
+target_include_directories (TimeFormatter PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+target_include_directories (FileAndFolderHandler PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+target_include_directories (Logging PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```
huntears, line 11: Why is there 3 different libraries just for the logging ?
That sounds excessive

logging/FileAndFolderHandler.cpp
```diff
@@ -0,0 +1,150 @@
...
+     * @brief Create the folder if it doesn't exist
+     *
+     * @param folder_name Name of the folder to create.
+     *
+     * @throw std::runtime_error if the folder can't be created
+     */
+    void FileAndFolderHandler::create_folder(std::string folder_name)
+    {
+        if (mkdir(folder_name.c_str(), 0777) != 0) {
+            switch(errno) {
```
huntears, line 19: Don't use a switch like that, just throw a runtime_error with `errno` translated to a string using `strerror`, that is enough and more readable

logging/FileAndFolderHandler.cpp
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @param filename Name of the file to check
+     *
+     * @return true The file exist
+     * @return false The file doesn't exist
+     */
+    bool FileAndFolderHandler::file_exist(std::string filename) const
+    {
+        std::fstream file;
+        file.open(filename);
```
huntears, line 106: Use stat here instead of opening the file

logging/FileAndFolderHandler.cpp
```diff
@@ -0,0 +1,150 @@
...
+        file.close();
+        return false;
+    }
+
+    /**
+     * @brief Get the Folder Path object
+     *
+     * @return std::string Path to the folder where log files are stored
+     */
+    std::string FileAndFolderHandler::get_folder_path() const
```
huntears, line 120: Please return a constant reference here

logging/FileAndFolderHandler.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        return this->_folder_path;
+    }
+
+    /**
+     * @brief Get the File Path object
+     *
+     * @return std::string Path to the current log file
+     */
+    std::string FileAndFolderHandler::get_file_path() const
```
huntears, line 130: Same here, return a constant reference

logging/FileAndFolderHandler_unittest.cpp
```diff
@@ -0,0 +1,79 @@
+#include <gtest/gtest.h>
+#include "FileAndFolderHandler.hpp"
+
+#include <cstdio>
+#include <unistd.h>
+
+namespace logging
```
huntears, line 7: For every tests here please create the files in `/tmp` as this would avoid any bad collisions

logging/Logger.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * @brief Write a message in the log file
+     *
+     * @param message Message to write
+     *
+     * @note The message will be written in the log file with the current time
+     */
+    void Logger::log(LogLevel level, std::string message)
```
huntears, line 43: Instead of having one function to log which takes a level in parameter, create one function per level, so that we can just call `logger.warn("A warning");` or `logger.error("An error");`, it would be much more readable

logging/Logger.hpp
```diff
@@ -0,0 +1,57 @@
...
+            void set_display_specification_level_in_console(LogLevel level);
+            void unset_display_specification_level_in_console(LogLevel level);
+            std::vector<LogLevel> get_display_specification_level_in_console() const;
+
+        private:
+            std::fstream _file_stream;                              /// Stream to the current log file
+            FileAndFolderHandler _file_and_folder_handler;          /// Handler for files and folders
+
+            std::vector<LogLevel> _specification_level_in_file;     /// Vector of LogLevel to display in the log file
+            std::vector<LogLevel> _specification_level_in_console;  /// Vector of LogLevel to display in the console
```
huntears, line 53: Please use `std::unordered_map` here instead of `std::vector` for both of those attributes

logging/Logger_unittest.cpp
huntears: No idea but it would be better to use function mocking to test the whole

logging/TimeFormatter.cpp
```diff
@@ -0,0 +1,95 @@
...
+        const std::tm* tm_now = std::localtime(&now);
+        int millis = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count() - (now * 1000);
+        std::stringstream ss;
+        ss << std::setw(2) << std::setfill('0') << tm_now->tm_year + 1900 << "/"
+           << std::setw(2) << std::setfill('0') << tm_now->tm_mon + 1 << "/"
+           << std::setw(2) << std::setfill('0') << tm_now->tm_mday << " "
+           << std::setw(2) << std::setfill('0') << tm_now->tm_hour << ":"
+           << std::setw(2) << std::setfill('0') << tm_now->tm_min << ":"
+           << std::setw(2) << std::setfill('0') << tm_now->tm_sec << "."
+           << std::setw(3) << std::setfill('0') << millis;
```
huntears, line 30: Why not use `std::put_time` ?

huntears (changes requested):
Some things to change but overall it is what we need, maybe a use of `std::filesystem` could be nice

Trompettesib:
> Some things to change but overall it is what we need, maybe a use of `std::filesystem` could be nice

Yes i know but before we were using c++ version without filesystem so I build the whole thing without it ^^

Trompettesib:
> > Some things to change but overall it is what we need, maybe a use of `std::filesystem` could be nice
> 
> Yes i know but before we were using c++ version without filesystem so I build the whole thing without it ^^

I think that it would be an upgrade in a future pull request but for the point this wednesday I will leave it like this

logging/CMakeLists.txt
Trompettesib: Because you said that it would be nice to split up all the code ^^
I will change that now

logging/FileAndFolderHandler_unittest.cpp
Trompettesib: how can I do that instead of hard coding the path ?

logging/Logger.hpp
Trompettesib: To store the "enum type" as a key and the string as a value ?

logging/TimeFormatter.cpp
Trompettesib: I didn't know this function ^^

logging/Logger_unittest.cpp
Trompettesib: Are you talking about that ? https://google.github.io/googletest/gmock_for_dummies.html
Because if yes, I didn't understand how can I do that

huntears (approved):
LGTM

PR #19, opened 2022-09-26
Packet callback system

huntears:
- [X] Parse packets from a client
- [x] Dispatch the packets to appropriate functions


Trompettesib (approved):
LGTM

PR #32, opened 2022-10-04
automation_and_tests_CD

Trompettesib:


Trompettesib:
I'm not sure about [this](https://github.com/CubicMC/cubic-server/blob/3322808c76c1b9fe2b745d8f954ae4a30512f3d0/.github/workflows/publish.yml#L48) part of the CD

huntears (changes requested):
There are still places where `/platform/bin/gcc` and `/platform/bin/g++` are used

huntears (approved):

PR #39, opened 2022-10-17
Add pyCraft for testing server packets

Trompettesib:


huntears (changes requested):
Please use a submodule for that

PR #41, opened 2022-11-03
Add ipv6 support

STMiki:


huntears:
This also works with ipv4 ?
It only handles ipv6 here from what I see

STMiki:
It works with both, ipv6 can handle ipv4 connection, we just need to enable it (on by default on linux)
`setsockopt(_sockfd, IPPROTO_IPV6, IPV6_V6ONLY, &no, sizeof(no));`

huntears (approved):

huntears (approved):

PR #42, opened 2022-11-08
World Data Structures

Trompettesib:


world/World.hpp
```diff
@@ -0,0 +1,29 @@
+#include <unordered_map>
+
+#include "LevelData.hpp"
+#include "Level.hpp"
+
+#ifndef WORLD_HPP
+#define WORLD_HPP
+
+class World
```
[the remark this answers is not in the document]
Trompettesib: In the World file from cubic-server/World.hpp it's "CUBICSERVER_WORLD_HPP" not "WORLD_HPP" but if you think it will be preferable to change it I will do it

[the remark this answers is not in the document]
STMiki: Well, what's the difference between the two ? How are they gonna be used ? I mean, there is two world class here, so I am a bit confused.

The one in cubic server host the dimensions & entities
(btw, why are entites in the world and not in the dimension ?)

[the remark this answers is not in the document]
Trompettesib: I don't know too...
I think we need to talk about that quickly

cubic-server/world-storage/WorldStorage.hpp
```diff
@@ -0,0 +1,28 @@
+#include <unordered_map>
+#include "LevelData.hpp"
+#include "Level.hpp"
+
+#ifndef WORLD_WORLD_HPP
+#define WORLD_WORLD_HPP
+
+class WorldStorage
```
STMiki, line 8: Ok, this may be stupid but why does this need to be separated from the World and Dimension class ?
From what I see here, this is just a data class for storing the levelData, and the levels that are for the dimension

So why not storing this in the existing class ?

huntears: I think it would make sense, as there are many helper functions you can have in this.
And you can also later on add functions just to load WorldStorage from the disk and bind it to a World.
Which could allow blockless worlds (For example to implement a queue, see [NanoLimbo](https://github.com/Nan1t/NanoLimbo)).

STMiki (dismissed):

.gitignore
```diff
@@ -31,7 +31,8 @@
 *.out
 *.app
 
-build
+build*
+build_*
```
huntears, line 35: Why ?

cubic-server/world-storage/WorldStorage.hpp
```diff
@@ -0,0 +1,28 @@
+#include <unordered_map>
+#include "LevelData.hpp"
+#include "Level.hpp"
+
+#ifndef WORLD_WORLD_HPP
+#define WORLD_WORLD_HPP
```
huntears, line 6: WORLD_WORLD ?
Why not WORLD_STORAGE ?

huntears (changes requested):

.gitignore
Trompettesib: Because I have multiple build folder for each branch I want to test

Trompettesib:
Do I have to remove all things related to Entity in my ChunkColumn ? Because technically it is useless I think

.gitignore
Trompettesib: Ok now after some reflection and usage, I'm ok with you I will delete it

cubic-server/world-storage/Level.cpp
```diff
@@ -0,0 +1,31 @@
...
+{
+}
+
+void Level::addChunkColumn(_2d_pos pos, ChunkColumn &chunkColumn) {
+    _chunkColumns.insert({pos, chunkColumn});
+}
+
+void Level::addChunkColumn(_2d_pos pos) {
+    ChunkColumn chunkColumn;
+    _chunkColumns.insert({pos, chunkColumn});
```
STMiki, line 17: This will crash.
`_chunkColumns` is a map of `ChunkColumn &`, and the created `chunkColumn` will only be valid inside the `addChunkColumn` function.

STMiki (changes requested):

cubic-server/world-storage/Level.cpp
Trompettesib: well done ! Thank you very much !

Trompettesib: Sadly this does not repair the segfault I have but it prevent a new one that will appear just after ^^

STMiki: It seems this is because of the hash ? I don't know, will check tomorrow.

STMiki (approved):

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include "Entity.hpp"
+
+#ifndef CHUNKCOLUMN_HPP
+#define CHUNKCOLUMN_HPP
+
+#define CHUNK_2D_SIZE 16*16
+#define CHUNK_3D_SIZE CHUNK_2D_SIZE*16
+#define BIOME_3D_SIZE 4*4*4
+#define NB_OF_CHUNKS 20
```
huntears, line 14: Please use `constexpr` when defining constants instead of defines

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+struct _3d_pos {
+    int x;
+    int y;
+    int z;
+};
```
huntears, line 20: Is this supposed to be here ?
I think putting it at another place to be usable by the rest of the code would be quite useful

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+class ChunkColumn
+{
+public:
+    ChunkColumn();
+    ~ChunkColumn();
+
+    void updateBlock(_3d_pos pos, uint16_t id);
+    uint16_t getBlock(_3d_pos pos);
+    const std::array<uint16_t, CHUNK_3D_SIZE*NB_OF_CHUNKS> &getBlocks();
```
huntears, line 62: Mark this method constant as it is not modifying the class

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,107 @@
...
+    const std::vector<block_entity *> &getBlockEntities();
+
+    int64_t getTick();
+    void setTick(int64_t tick);
+
+    void updateEntity(std::size_t id, Entity *e);
+    void updateEntity(__int128 uuid, Entity *e);
+    void addEntity(Entity *e);
+    void removeEntity(std::size_t id);
+    void removeEntity(__int128 uuid);
```
huntears, line 89: Instead of using `__int128` you should make a `typedef` to specify what it is.
It will also allow us to support other compilers other than gcc down the line

cubic-server/world-storage/ChunkColumn.cpp
```diff
@@ -0,0 +1,123 @@
+#include "ChunkColumn.hpp"
+#include <memory>
+
+ChunkColumn::ChunkColumn()
```
huntears, line 4: Use `default` instead of creating an empty constructor

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,107 @@
...
+    void updateHeightMap(void);
+    const height_map &getHeightMap(void);
+private:
+    std::array<uint16_t, CHUNK_3D_SIZE*NB_OF_CHUNKS> _blocks;
+    std::array<uint8_t, CHUNK_3D_SIZE*NB_OF_CHUNKS> _skyLights;
+    std::array<uint8_t, CHUNK_3D_SIZE*NB_OF_CHUNKS> _blockLights;
+    std::array<uint8_t, BIOME_3D_SIZE*NB_OF_CHUNKS> _biomes;
+    std::vector<block_entity *> _blockEntities;
+    int64_t _tickData;
+    std::deque<Entity *> _entities;
```
huntears, line 103: Why is this a `std::deque` instead of a `std::vector` ?

cubic-server/world-storage/Level.hpp
```diff
@@ -0,0 +1,47 @@
...
+};
+
+template<>
+struct std::hash<_2d_pos>
+{
+    std::size_t operator()(_2d_pos const& pos) const noexcept
+    {
+        std::size_t h1 = std::hash<int>{}(pos.x);
+        std::size_t h2 = std::hash<int>{}(pos.z);
+        return h1 ^ (h2 << 1);
```
huntears, line 25: This sounds like collisions but I am not sure.

cubic-server/world-storage/Level.cpp
```diff
@@ -0,0 +1,33 @@
+#include "Level.hpp"
+
+Level::Level()
```
huntears, line 3: Use `default` here

cubic-server/world-storage/Level.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+}
+
+Level::~Level()
+{
+}
+
+ChunkColumn *Level::addChunkColumn(_2d_pos pos, ChunkColumn *chunkColumn) {
+    _chunkColumns.insert({pos, chunkColumn});
+    return _chunkColumns.at(pos);
```
huntears, line 13: Instead of doing an `insert` with an `at` afterwards, `insert` already returns an iterator to the object inserted and the status of the insertion.

cubic-server/world-storage/Level.cpp
```diff
@@ -0,0 +1,33 @@
...
+    _chunkColumns.insert({pos, new ChunkColumn()});
+    return _chunkColumns.at(pos);
+}
+
+ChunkColumn *Level::getChunkColumn(_2d_pos pos) {
+    return _chunkColumns.at(pos);
+}
+
+ChunkColumn *Level::getChunkColumn(int x, int z) {
+    _2d_pos pos = {x < 0 ? -1 + int(x/16) : int(x/16), x < 0 ? -1 + int(z/16) : int(z/16)};
```
huntears, line 26: No

cubic-server/world-storage/WorldStorage.hpp
```diff
@@ -0,0 +1,29 @@
...
+public:
+    WorldStorage();
+    ~WorldStorage();
+
+    const LevelData &getLevelData() const;
+    void setLevelData(const LevelData &value);
+
+    const std::vector<std::string> getLevelNames() const;
+    const Level *getLevel(std::string levelName) const;
+    Level *getLevelM(std::string levelName);
```
huntears, line 19: Please use a constant reference for arguments with objects, especially when it is a `std::string`

huntears: Also please rename this method, for example `getEditableLevel`

huntears (changes requested):
- All the code in the folder `world-storage` should be in a namespace (This would avoid naming conflict that could arise)

cubic-server/world-storage/Level.hpp
Trompettesib: what do you mean by collisions ?

cubic-server/world-storage/Level.cpp
Trompettesib: do you want me to do something like that ?
```cpp
return _chunkColumns.insert({pos, chunkColumn}).first->second;
```

cubic-server/world-storage/Level.cpp
Trompettesib: YES

Trompettesib:
I have removed entities from chunk columns because I think we don"t need it every were

cubic-server/world-storage/ChunkColumn.hpp
```diff
@@ -0,0 +1,83 @@
...
+constexpr int BIOME_3D_SIZE = 4*4*4;
+constexpr int NB_OF_CHUNKS = 20;
+
+struct block { // TODO: PTDR
+};
+
+struct block_entity {
+};
+
+struct height_map {
```
huntears, line 23: Can you name all the structs in PascalCase like the classes please ?

huntears (changes requested):

cubic-server/world-storage/ChunkColumn.hpp
Trompettesib: 👍

cubic-server/world-storage/ChunkColumn.hpp
Trompettesib: Because deque is more suitable to big change in size (from what I read)

huntears (approved):

STMiki (approved):
👌

PR #43, opened 2022-11-08
Refactor to the new architecture

huntears:
This also refactors the logger and the configuration handling.


huntears:
I just need to merge master into it, and it should be good

STMiki (approved):

cubic-server/Server.hpp
```diff
@@ -13,98 +13,50 @@
...
 
 class Server
 {
 public:
-    Server();
     ~Server();
 
     void launch();
 
+    const Configuration::ConfigHandler &getConfig() const {
```
Trompettesib, line 31: Should we have to put simple getters in hpp or should we put them in the corresponding cpp ?

Trompettesib (approved):
Look good to me !
Maybe in the future we should think about new subfolders if we add more files

PR #44, opened 2022-11-10
Update ci to github-hosted runners

Trompettesib:


huntears (commented):
Wouldn't it be better to put the installation of the packages before checking out the repo ?

huntears (approved):

PR #48, opened 2022-11-15
Chat

STMiki:
- [x] Chat class
- [x] Message class
- [x] Way to call the Chat from the player
  (Player -> Dimension -> World -> Chat)
- [ ] Way to filter client depending on chat visibility
- [x] Way to get client depending on their world group
  (for now, we just get all clients)

Functions
- [x] sendPlayerMessage
  (All player message are wrapped in a translation component. This is done here, then sent to all players. I don't know if there is more component for the other function, maybe for the say or msg command)
- [x] sendSystemMessage
- [x] sendSayMessage
- [x] sendMsgMessage
- [ ] sendTeamMessage
- [ ] sendEmoteMessage (Yes. I don't know either...)
- [ ] sendTellrawMessage


STMiki:
I changed how Dimension are stored in the World class. I needed to access them from the World, and they were initialized/stored in the derived class.

cubic-server/Chat.hpp
```diff
@@ -1,11 +1,133 @@
...
 
+#include <stdint.h>
+#include <string>
+#include "Logger.hpp"
+#include "nlohmann/json.hpp"
+
+class Player;
+class Client;
+
+enum class MsgType : int32_t {
```
STMiki, line 12: Maybe move this inside the class now

cubic-server/Client.hpp
```diff
@@ -61,6 +62,20 @@ class Client
     // All the send packets go here
     void sendStatusResponse(const std::string &json);
     void sendPingResponse(int64_t payload);
+    // Should I just pass the packet directly?
```
Trompettesib, line 65: I think so, it's gonna be more readable

protocol/ServerPackets.hpp
```diff
@@ -138,6 +139,19 @@ namespace protocol
     };
     std::shared_ptr<ChangeDifficulty> parseChangeDifficulty(std::vector<uint8_t> &buffer);
 
+    struct ChatMessage : BaseServerPacket
+    {
+        std::string message;
+        // I think this is a string, it is marked as an 'Instant' in the protocol
```
Trompettesib, line 145: I think this is a long like in the "chat command" packet

Trompettesib (dismissed):
LGTM but let @huntears review it to be sure

You're just insane !

protocol/ServerPackets.hpp
STMiki: https://docs.oracle.com/javase/8/docs/api/serialized-form.html#java.time.Instant
Ok, we are both wrong x)
I will do that tomorrow

cubic-server/Chat.cpp
```diff
@@ -1 +1,267 @@
...
+    }
+    const auto clients = Server::getInstance()->getClients();
+
+    nlohmann::json response;
+    response["translate"] = "chat.type.text";
+    response["with"] = nlohmann::json::array();
+    response["with"].push_back({"text", "PlayerName"});
+    response["with"].push_back(message.toJson());
+
+    // TODO: Filter clients by WorldGroup & chat visibility.
```
STMiki, line 25: Wait, I can do that now...

Trompettesib (dismissed):
LGTM

STMiki:
I replaced the try catch for a null check, because I am not working with reference anymore, and dynamic_cast only throw when working with ref

STMiki:
> LGTM

I swear that's the last change ! :laughing:

Trompettesib (approved):
SLGTM (Still looking good to me) ^^
This is way better without the try catch !

protocol/ClientPackets.cpp
```diff
@@ -21,4 +21,24 @@ std::shared_ptr<std::vector<uint8_t>> protocol::createStatusResponse(const Statu
...
+    serialize(payload, in.unsignedContent, addString);
+    serialize(payload, in.type, addVarInt);
+    serialize(payload, in.senderUUID, addUUID);
+    serialize(payload, in.senderName, addString);
+    serialize(payload, in.hasTeamName, addBoolean);
+    serialize(payload, in.teamName, addString);
+    serialize(payload, in.timestamp, addLong);
+    serialize(payload, in.salt, addLong);
+    serialize(payload, in.signatureLength, addVarInt);
+    serialize(payload, in.signature, addByteArray);
```
huntears, line 40: Please only use one call to `serialize`, it has a variable number of arguments like `parse`

huntears (changes requested):
Overall really good PR, just one thing to change and I would consider it perfect 👍

huntears (approved):

huntears:
Perfect PR, I love it :D

PR #51, opened 2022-11-17
Fix launch of tests

Trompettesib:


huntears (changes requested):
Why are you going into build while it is already doing that with `working-directory` ?

Trompettesib:
> Why are you going into build while it is already doing that with `working-directory` ?

Yes I know 😞

huntears (approved):

PR #52, opened 2022-11-18
Disconnect player

LEVRAIBHL:


cubic-server/Client.cpp
```diff
@@ -310,3 +310,13 @@ void Client::sendPingResponse(int64_t payload)
 
     LDEBUG("Sent a ping response");
 }
+
+void Client::sendDisconnectPlayer(Chat reason)
+{
+    auto pck = protocol::disconnectPlayer({
```
STMiki, line 316: Since #48 is merged, you can replace that with `chat::Message`.

huntears (changes requested):
Please update your branch on master

PR #53, opened 2022-11-18
Disconnect player

LEVRAIBHL:


cubic-server/Client.cpp
```diff
@@ -293,6 +293,18 @@ void Client::_onEncryptionResponse(const std::shared_ptr<protocol::EncryptionRes
     _log->debug("Got a Encryption Response");
 }
 
+void Client::_onDisconnectRequest(const std::shared_ptr<protocol::DisconnectReason> &pck, chat::Message &message)
```
Trompettesib, line 296: This function is great but you don't need to use
```cpp
const std::shared_ptr<protocol::DisconnectReason> &pck
```
and there isn't any request of disconnection. Maybe you should name your function like this :
```cpp
void Client::DisconnectClient(chat::Message &message)
```

Trompettesib (changes requested):
Looks good ! few necessary changes

Trompettesib (dismissed):
LGTM

cubic-server/Client.cpp
```diff
@@ -320,3 +332,13 @@ void Client::sendChatMessageResponse(const protocol::PlayerChatMessage &packet)
...
 }
+
+void Client::sendDisconnectPlayerReason(const std::string &json)
+{
+    auto pck = protocol::DisconnectPlayerReason({
+        json
+    });
+    _sendData(*pck);
+
+    LDEBUG("Sent a disconnet reason for player");
```
huntears, line 343: The wording here is a bit weird, why not something like
`Sent a disconnect player reason`
That would be coherent with the rest of the debug logs here.

huntears (changes requested):
Some small things to change, but overall good.

Trompettesib (approved):
LGTM

cubic-server/protocol/ClientPackets.hpp
```diff
@@ -47,7 +48,13 @@ namespace protocol
     };
 
     std::shared_ptr<std::vector<uint8_t>> createPlayerChatMessage(const PlayerChatMessage &);
+    
+    struct DisconnectReason
+    {
+        std::string reason;
+    };
 
+    std::shared_ptr<std::vector<uint8_t>> DisconnectPlayerReason(const DisconnectReason &);
```
huntears, line 57: I missed that on the first review, but this should be named like the rest of the functions here.

For example `createDisconnectPlayerReason`

huntears (changes requested):

Trompettesib (approved):
👌

cubic-server/Client.hpp
```diff
@@ -75,6 +76,7 @@ class Client
     void _onLoginStart(const std::shared_ptr<protocol::LoginStart> &pck);
     void _onPingRequest(const std::shared_ptr<protocol::PingRequest>& pck);
     void _onEncryptionResponse(const std::shared_ptr<protocol::EncryptionResponse> &pck);
+    void disconnectClient(chat::Message &message);
```
huntears, line 79: Just realized that this should be a public method, not a private.

huntears (changes requested):

Trompettesib (approved):
👌

cubic-server/Client.cpp
```diff
@@ -293,6 +293,16 @@ void Client::_onEncryptionResponse(const std::shared_ptr<protocol::EncryptionRes
     _log->debug("Got a Encryption Response");
 }
 
+void Client::disconnectClient(chat::Message &message)
+{
+    nlohmann::json json;
+    
+    json["translate"] = "chat.type.text";
+    json["with"] = nlohmann::json::array({{"text", "PlayerName"}, {message.toJson()}});
```
huntears, line 301: Please go on multiple lines here

huntears (changes requested):

cubic-server/Client.cpp
STMiki: That's probably my fault, this is supposed to be in the Message class...

cubic-server/Client.hpp
```diff
@@ -65,6 +65,8 @@ class Client
     void sendPingResponse(int64_t payload);
     // Should I just pass the packet directly?
     void sendChatMessageResponse(const protocol::PlayerChatMessage &packet);
+    void disconnectClient(chat::Message &message);
```
huntears, line 68: `message` should be a constant reference

huntears (changes requested):

Trompettesib (commented):
Same as huntears, after that I think this would be perfect !

STMiki (dismissed):
👍

Trompettesib (dismissed):
LGTM

huntears (changes requested):

cubic-server/Client.hpp
```diff
@@ -65,6 +65,8 @@ class Client
     void sendPingResponse(int64_t payload);
     // Should I just pass the packet directly?
     void sendChatMessageResponse(const protocol::PlayerChatMessage &packet);
+    void disconnectClient(const chat::Message &message);
```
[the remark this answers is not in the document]
huntears: And maybe also just call the function `disconnect` instead of `disconnectClient`, since the `Client` would be repetitive.
You could also give a default reason if none is passed like `You have been disconnected by the server` for example.

PR #54, opened 2022-11-28
Entity system

Trompettesib:


cubic-server/Dimension.hpp
```diff
@@ -19,8 +23,12 @@ class Dimension
     virtual void initialize();
     virtual void tick();
     virtual World *getWorld() const;
+    virtual std::vector<Entity *> getEntities();
+    virtual void forEachEntity(std::function<void(Entity *)> callback);
+    virtual void forEachEntityIf(std::function<bool(Entity *)> predicate, std::function<void(Entity *)> callback);
```
huntears, line 28: Please put the callback first

cubic-server/Dimension.hpp
```diff
@@ -19,8 +23,12 @@ class Dimension
     virtual void initialize();
     virtual void tick();
     virtual World *getWorld() const;
+    virtual std::vector<Entity *> getEntities();
```
huntears, line 26: You sure you want a copy here ?

cubic-server/Entity.hpp
```diff
@@ -4,18 +4,53 @@
...
+        SNEAKING,
+        LONG_JUMPING,
+        DYING,
+        CROAKING,
+        USING_TONGUE,
+        ROARING,
+        SNIFFING,
+        EMERGING,
+        DIGGING
+    };
```
huntears, line 27: Please don't put then in uppercase like this.
Just use PascalCase

cubic-server/Entity.cpp
```diff
@@ -1,5 +1,27 @@
...
+    _airTicks = 300;
+    _customName = "";
+    _customNameVisible = false;
+    _silent = false;
+    _noGravity = false;
+    _pose = Pose::STANDING;
+    _tickFrozenInPowderedSnow = 0;
+}
+
+Entity::~Entity()
```
huntears, line 21: The destructor has been deleted

cubic-server/World.cpp
```diff
@@ -32,5 +32,21 @@ std::shared_ptr<Dimension> World::getDimension(const std::string_view &name) con
 
 std::vector<Entity *> World::getEntities()
 {
-    return _entities;
+    std::vector<Entity *> entities;
+    for (auto & _dimension : _dimensions)
+        for (auto & _entity : _dimension.second->getEntities())
+            entities.push_back(_entity);
+    return entities;
 }
```
huntears, line 40: Caching that would be really nice.

cubic-server/Entity.cpp
Trompettesib: I know but if I remove that the linker cannot link for a reason I don't understand
image

STMiki: https://stackoverflow.com/a/1219618

> Note: The destructor is the only method that even if it is pure virtual has to have an implementation in order to instantiate derived classes (yes pure virtual functions can have implementations).

huntears (approved):

STMiki (approved):
👍

PR #63, opened 2022-12-13
Add textbox + event handler (key not bind)

NekoSelena:
The 'Enter' key binding is not handled, but it works with other keys. It's an issue to be fixed while handling the commands parsing (to get the user's input).

### DIFF
--- a/cubic-server/interface/LogsInterface.cpp
+++ b/cubic-server/interface/LogsInterface.cpp
@@ -6,21 +6,61 @@
 //#include "Logger.hpp"
 #include "LogsInterface.hpp"
 
-LogsInterface::LogsInterface()
+LogsInterface::LogsInterface() :
+  m_VBox(Gtk::ORIENTATION_VERTICAL)
 {
     set_shadow_type(Gtk::SHADOW_ETCHED_OUT);
     set_policy(Gtk::POLICY_AUTOMATIC, Gtk::POLICY_AUTOMATIC);
-    add(m_label);
+    add(m_VBox);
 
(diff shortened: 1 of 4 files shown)


cubic-server/interface/LogsInterface.cpp
```diff
@@ -6,23 +6,63 @@
...
+{
+  if (event->keyval == GDK_KEY_KP_Enter)
+  {
+    // Supposed to add the input to the logs (interface side only) when 'ENTER'
+    // is pressed. /!\ To do when handling command parsing.
+    m_label.set_text(m_label.get_text() + "\n" + m_Entry.get_text());
+    m_Entry.set_text("");
+    return true;
+  }
+  else if(event->keyval == GDK_KEY_Escape)
```
Trompettesib, line 55: The else here is useless I think

cubic-server/interface/LogsInterface.hpp
```diff
@@ -9,8 +9,13 @@ class LogsInterface : public Gtk::ScrolledWindow {
     virtual ~LogsInterface();
     bool on_timeout();
 
+    private:
```
Trompettesib, line 12: Indentation are not respected here

cubic-server/interface/LogsInterface.cpp
huntears: That is not a problem tbh

Trompettesib (approved):
LGTM

huntears (approved):

PR #67, opened 2022-12-20
Add NBT types + serialization

huntears:
It should just work.
The only thing that could be made better later down the line would be `nbt::List`, it only throws an exception for multiple types at serialization, not during initialization.


huntears:
Some problems are in this PR, currently fixing.

Trompettesib:
Testing it with google test would have been perfect

Trompettesib (approved):
LGTM

Trompettesib (approved):
👌

STMiki (approved):
👌

PR #69, opened 2022-12-22
Sound system

STMiki:
The tick function in the `SoundSystem` is only here for testing purposes.


huntears (approved):
It looks fine, but we will surely need to re-architecture this part at the end of this sprint.
(Like most of the new code pepega)

Trompettesib (approved):
👌 👍

PR #71, opened 2022-12-25
Fix logs update (commented for some reason)

NekoSelena:
A part of the code was commented with no apparent reason after merging Logs Interface PR into master


Trompettesib (approved):
LGTM

STMiki (approved):
Looks good, and I think what you are looking to do [here](https://github.com/CubicMC/cubic-server/blob/fix-logs-interface/cubic-server/interface/LogsInterface.cpp#L43) is catch the `activate` signal.

huntears (approved):

PR #72, opened 2023-01-02
Add argument parsing

NekoSelena:


Trompettesib (dismissed):
LGTM
Just the global boolean is maybe to change but I don't know how

huntears:
It would be better if the argument parsing was done by a proper library, so we can generate a usage output and add more complex options in the future.
Something like `https://github.com/p-ranav/argparse` would be quite nice.

cubic-server/main.cpp
```diff
@@ -1,32 +1,45 @@
 #include <iostream>
 #include <cstring>
+#include <unordered_map>
```
Trompettesib, line 3: These includes are not longer necessary (unordered map and functional)

cubic-server/main.cpp
```diff
@@ -1,32 +1,45 @@
...
+#include <argparse/argparse.hpp>
 #include "Server.hpp"
 #include "protocol/ServerPackets.hpp"
 #include "interface/ManagementInterface.hpp"
 
-static void print_usage(const char *caller)
-{
-    std::cout << "Usage:\n\t" << caller << std::endl;
-}
+bool arg_parser(int argc, char **argv) {
```
Trompettesib, line 11: Maybe you should change the name of the function to camelCase

Trompettesib (dismissed):
Great !

cubic-server/main.cpp
```diff
@@ -1,32 +1,45 @@
...
+        std::cerr << program;
+        std::exit(1);
+    }
 
-// Uncomment that to launch the interface
-// #define INTERFACE
+    if (program["--nogui"] == false) {
+        return false;
+    }
+    return true;
```
STMiki, line 30: For now this works because we only check the gui, but if we add more options later, it would be best to return a map of arguments, or the argument parser itself.

huntears (approved):
Code looks great and testing on my side went smoothly, everything is good 👍

huntears:
I think having a proper version being shown with `--version` would be nice, but it isn't a requirement.
I will add it to an issue ig

STMiki (approved):
👌

PR #75, opened 2023-01-02
Add command parsing

Trompettesib:


Trompettesib:
When I try to use it in the main for testing, I have a linker error that says "multiple definition" and I don't know how to solve that

Trompettesib:
It was compiling on my machine even if the include was misspelled IDK why

NekoSelena (dismissed):

cubic-server/CMakeLists.txt
```diff
@@ -47,3 +47,4 @@ add_subdirectory (configuration)
 add_subdirectory (logging)
 add_subdirectory (interface)
 add_subdirectory (world_storage)
+add_subdirectory (command-parser)
```
STMiki, line 50: command_parser

cubic-server/command-parser/CommandParser.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef CUBICSERVER_COMMANDPARSER_HPP
+#define CUBICSERVER_COMMANDPARSER_HPP
+
+#include <string>
+#include <unordered_map>
+#include <functional>
+
+#include "commands/help.hpp"
+
+namespace CommandParser {
```
STMiki, line 10: command_parser
We changed it for world_storage

cubic-server/command-parser/commands/help.cpp
```diff
@@ -0,0 +1,23 @@
+#include "help.hpp"
+
+namespace help {
+
+    void help(std::vector<std::string>& args) {
+        logging::Logger::get_instance()->info("/help [<command>]");
```
STMiki, line 6: Why are you not calling the macro ?

STMiki: Ok, so I just saw how the macro works, we need to change that I think. x)

cubic-server/command-parser/commands/help.hpp
```diff
@@ -0,0 +1,100 @@
...
+    void help(std::vector<std::string>& args);
+    void execute(std::vector<std::string>& args);
+
+    //TODO: Move this code
+    static const std::unordered_map<std::string, std::function<void(std::vector<std::string>&)>> helps {
+            {"help", help::help}
+    };
+
+    static const std::vector<std::string> allCommands = {
+        "/advancement (grant|revoke)",
```
STMiki, line 22: Can plugins add commands ?
If so, we need a way to add them to the help command too. It would be better to have a command class that can be added to a registry or something, event if they just contain static function.

Trompettesib: Do you mean something like that ? 
```cpp
namespace help_test {

struct helptest
{
    static void help(std::vector<std::string>& args);
    static void execute(std::vector<std::string>& args);
    inline static const std::string name = "/help [<command>]";
};

static const std::unordered_map<std::string, std::function<void(std::vector<std::string>&)>> helps {
    {"help", help_test::helptest::help},
    {"?", help_test::helptest::help}
};

static const std::vector<std::string> allCommands = {
    help_test::helptest::name,
};
}
```

Trompettesib: I still need to work on a more reliable way of storing these map/vector 'cause each function will have to be included in the help.hpp to be in these map/vector

STMiki: Not really. Basically, we need a base command class with at least 3 virtual commands : `execute`, `help` and `autocomplete`.

For now, we just need `execute` and `help`. The will do the same as what you did, but the advantage, is that we can store commands, add some (via plugins for example), override some, etc...

This is useful in your case, because you can see what command is stored, and create the help message with only those.

huntears (approved):

cubic-server/command-parser/commands/help.cpp
huntears: We will change that later, not really a big deal right now

NekoSelena (approved):

PR #76, opened 2023-01-07
Fix NBT destructors

huntears:
An NBT object will now only get destroyed when the proper function is called

### DIFF
--- a/cubic-server/nbt.hpp
+++ b/cubic-server/nbt.hpp
@@ -185,6 +185,7 @@ public:
 
     constexpr void destroy() override {
         for (auto i : _value) {
+            i->destroy();
             delete i;
         }
     }
@@ -440,6 +441,7 @@ public:
 
     constexpr void destroy() override {
         for (auto i : _value) {
+            i->destroy();
             delete i;
         }
     }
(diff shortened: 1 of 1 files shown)


Trompettesib (approved):
👌

Trompettesib:
🙏

cubic-server/nbt.hpp
```diff
@@ -183,6 +182,12 @@ class Compound : public Base
         // Ends the TAG_Compound with a TAG_End
         data.push_back(0);
     }
+
+    constexpr void destroy() override {
+        for (auto i : _value) {
+            delete i;
```
STMiki, line 188: I think you need to call destroy before deleting it, cause if you have a nested list or compound, it will not be freed

cubic-server/nbt.hpp
```diff
@@ -432,6 +437,12 @@ class List : public Base
             i->serialize(data, false);
         }
     }
+
+    constexpr void destroy() override {
+        for (auto i : _value) {
+            delete i;
```
STMiki, line 443: Same

STMiki (changes requested):

Trompettesib (approved):

STMiki (approved):
Noice

PR #88, opened 2023-01-15
Make the server not commit sepuku every tick

huntears:
This makes the server not create 3 threads every tick

### DIFF
--- a/cubic-server/Dimension.hpp
+++ b/cubic-server/Dimension.hpp
@@ -6,6 +6,7 @@
 #include <thread>
 #include <vector>
 #include <functional>
+#include <semaphore>
 
 #include "logging/Logger.hpp"
 #include "Entity.hpp"
@@ -18,7 +19,7 @@ class Entity;
 class Dimension
 {
 public:
-    Dimension(World *world): _world(world) {
+    Dimension(World *world): _world(world), dimensionLock(std::counting_semaphore<1000>(0)) {
         _log = logging::Logger::get_instance();
     }
     virtual void initialize();
@@ -31,6 +32,7 @@ public:
     const world_storage::Level &getLevel() const;
     world_storage::Level &getEditableLevel();
     virtual void generateChunk(int x, int z);
+    std::counting_semaphore<1000> dimensionLock;
 
 protected:
     std::vector<Entity *> _entities;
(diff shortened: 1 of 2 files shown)


cubic-server/Dimension.hpp
```diff
@@ -18,7 +19,7 @@ class Entity;
 class Dimension
 {
 public:
-    Dimension(World *world): _world(world) {
+    Dimension(World *world): _world(world), dimensionLock(std::counting_semaphore<1000>(0)) {
```
Trompettesib, line 22: 1000 is a magical value here, please use a define or a constexpr

Trompettesib (approved):

cubic-server/default/DefaultWorld.cpp
```diff
@@ -6,14 +6,21 @@ DefaultWorld::DefaultWorld(WorldGroup *worldGroup)
     this->_dimensions.emplace("end", std::make_shared<TheEnd>(reinterpret_cast<World *>(this)));
     this->_dimensions.emplace("nether", std::make_shared<TheNether>(reinterpret_cast<World *>(this)));
     this->_dimensions.emplace("overworld", std::make_shared<Overworld>(reinterpret_cast<World *>(this)));
+    for (auto [_, dim] : this->_dimensions)
+        _processingThreads.push_back(new std::thread([dim](){
+            while (1) {
+                dim->dimensionLock.acquire();
+                dim->tick();
```
STMiki, line 13: This will not run at 20t/s

STMiki: Btw, do you need to release the lock ? I am not familiar with this type

STMiki (changes requested):

cubic-server/default/DefaultWorld.cpp
huntears: It is running at 20 ticks per second, the semaphore is released by the global tick rate. Therefore it will not run faster than 20 ticks per seconds, except if the computation is lagging behind, which would make it not wait for the 50ms to start computing

STMiki: big brain move

STMiki (approved):

PR #106, opened 2023-02-01
1.19.3 support

huntears:
That's gonna be a lot of pain, but here is the list of things to do (Will get updated when I find new stuff) : 
- [x] Player Info Handling (They somehow changed how it changed entirely)
- [x] Update all client bound packet IDs
- [x] Update all server bound packet IDs
- [ ] Check what happened with `Custom Sound Effect` (Disappeared in the latest wiki.vg revision)
- [x] Check that everything works as expected in-game
- [x] Update README
- [ ] Fix loading terrain taking a billion years to end
- [x] Fix the protocol IDs

### DIFF
--- a/cubic-server/Player.cpp
+++ b/cubic-server/Player.cpp
@@ -371,40 +371,47 @@ void Player::sendLoginPlay(const protocol::LoginPlay &packet)
     getDimension()->spawnPlayer(this);
 
     // Send login message
-    chat::Message connectionMsg = chat::Message("", {
-        .color = "yellow",
-        .translate = "multiplayer.player.joined",
-        .with = std::vector<chat::Message>({
-            chat::Message(
-                this->getUsername(),
-                {
-                    .insertion = this->getUsername(),
-                },
-                chat::message::ClickEvent(
-                    chat::message::ClickEvent::Action::SuggestCommand,
-                    "/tell " + this->getUsername() + " "
-                ),
-                chat::message::HoverEvent(
-                    chat::message::HoverEvent::Action::ShowEntity,
-                    "{\"type\": \"minecraft:player\", \"id\": \"" + this->getUuidString() + "\", \"name\": \"" + this->getUsername() + "\"}"
-                )
-            )
-        })
-    });
+    // chat::Message connectionMsg = chat::Message("", {
+    //     .color = "yellow",
+    //     .translate = "multiplayer.player.joined",
+    //     .with = std::vector<chat::Message>({
+    //         chat::Message(
+    //             this->getUsername(),
+    //             {
+    //                 .insertion = this->getUsername(),
+    //             },
+    //             chat::message::ClickEvent(
+    //                 chat::message::ClickEvent::Action::SuggestCommand,
+    //                 "/tell " + this->getUsername() + " "
+    //             ),
+    //             chat::message::HoverEvent(
+    //                 chat::message::HoverEvent::Action::ShowEntity,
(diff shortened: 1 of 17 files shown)


huntears:
Custom Sound Effect is possibly merged into another packet, I'll check that now

huntears:
Yeah, I think it has been merged into the Sound Effect packet.
See https://pokechu22.github.io/BurgerWebViewer/#packets:play_clientbound_5e
Though I have no idea how that packet is supposed to work since it has not yet been documented on wiki.vg.

I'll try to get something out of it but that's not promising :/

huntears:
Currently I don't have many leads on why the terrain loading takes so long.
My guess is that the missing chat information in the registry codec might be causing this, but aight idrk

huntears:
Loading terrain still takes forever and Custom Sound Effect is not implemented but it should work for everything else.

Trompettesib (dismissed):
LGTM

cubic-server/Player.cpp
```diff
@@ -345,11 +346,42 @@ void Player::sendBlockUpdate(const protocol::BlockUpdate &packet)
...
     auto pck = protocol::createLoginPlay(packet);
     _cli->_sendData(*pck);
     LDEBUG("Sent a login play");
+
+    this->sendFeatureFlags({
+        {"minecraft:vanilla"}
+    });
+
+    this->sendPlayerAbilities({
+        4 | 1 | 2 | 8,
```
STMiki, line 372: Comment, wtf are theses

cubic-server/World.cpp
```diff
@@ -132,25 +134,47 @@ void World::updateTime() {
 void World::sendPlayerInfoAddPlayer(Player *current) {
     // get list of players
     std::vector<Player *> players = this->getPlayers();
-    std::vector<protocol::_Player> players_info;
+    std::vector<protocol::_Actions> players_info;
+
+    uint8_t actions = 0x3f; // TODO: bruh
```
STMiki, line 139: Same

cubic-server/protocol/ClientPackets.cpp
```diff
@@ -389,68 +408,77 @@ std::shared_ptr<std::vector<uint8_t>> protocol::createSystemChatMessage(const Sy
...
-                if (player.updateDisplayName.hasDisplayName)
-                    serialize(payload, player.updateDisplayName.displayName, addChat);
-                break;
-
-            default:
-                break;
+    for (auto &actionSet : in.actionSets) {
+        serialize(payload, actionSet.uuid, addUUID);
+
+        if (in.actions & 0b00000001) { // Add Player
```
STMiki, line 435: A enum would be cool

cubic-server/protocol/serialization/addPrimaryType.hpp
```diff
@@ -185,6 +185,11 @@ namespace protocol
     {
         addLong(out, ((data.x & 0x3FFFFFF) << 38) | ((data.z & 0x3FFFFFF) << 12) | (data.y & 0xFFF));
     }
+
+    constexpr void addBitSet(std::vector<uint8_t> &out, const std::vector<uint64_t> &bitset)
```
STMiki, line 189: Why not just use `addArray<uint64_t, addUnsignedLong>` ?

This can be confusing because we now have the `popBitSet` which return an `std::bitset`

huntears: That was old code, it isn't getting executed.
Getting rid of it

STMiki (approved):
👌

Trompettesib (approved):

PR #108, opened 2023-02-09
Block state code generation

Trompettesib:


generators/final.hpp
```diff
@@ -0,0 +1,199 @@
...
+        case 7050:
+        case 7051:
+        case 7052:
+        case 7053:
+        case 7054:
+        case 7055:
+        case 7056:
+        case 7057:
+        case 7058:
+            return "minecraft::acacia_button";
```
STMiki, line 195: `minecraft:acacio_button`

STMiki (commented):
Add the blocks.json from mc to the gitignore.

Trompettesib:
I think I have to concat all the toName function but I'm not sure about that

STMiki (dismissed):
Looks good, but does a test compile ?

Trompettesib:
> Looks good, but does a test compile ?

nop because of the toName functions

Trompettesib:
> Add the blocks.json from mc to the gitignore.

Do I need to delete it on github ?

Trompettesib:
Now it's compiling !

STMiki:
> > Add the blocks.json from mc to the gitignore.
> 
> Do I need to delete it on github ?

I think so ? but this will be generated from the

STMiki:
Oops

STMiki (commented):
Mhh.. I don't know how to go about this.

Either let the generated.hpp file here,
or add it to the gitignore and generate it from the cmake.

But if we generate it, we need to download the .jar, extract the data and generate the .hpp

🤷

Trompettesib:
I have a question.
In the function toName, do I have to return a std::pair<std::string, std::vector<std::pair<std::string, std::string>>> ?
Like, I'm only returning the name of the block for any protocol id it has. Maybe it should be better to return the name, and the properties if the block has some ? as in the fromNameToProtocolId function

STMiki:
I think you should hide these in a block class

generators/protocol_block_ids.py
```diff
@@ -0,0 +1,275 @@
...
+            data += "case " + str(state["id"]) + ":\n"
+            data += "return {\"" + self.name + "\", {"
+            if self.properties != []:
+                for prop in self.properties:
+                    data += "{\"" + prop + "\", \"" + state["properties"][prop] + "\"}, "
+                data = data[:-2]
+            data += "}};\n"
+        return data
+
+blocks = []
```
STMiki, line 192: Put theses in a function, because this script could be called from another one idk.

STMiki: call it from `if __name__ == "__main__":` thingy

STMiki (commented):
👌

STMiki (approved):
👌

huntears:
I also forgot but documenting the generator a little bit would be nice ;)

generators/protocol_block_ids.py
```diff
@@ -264,25 +250,54 @@ def write_header_bis(blocks):
...
+
+# this is the main function
 def main():
-    blocks = load_json("blocks.json")
-    write_header(blocks)
-    write_header_bis(blocks)
+    options = {"input": "blocks.json", "output": "generated/blocks"}
+    options = parse_args(options)
+    blocks = load_json(options["input"])
+    os.makedirs('/'.join(options["output"].split('/')[:-1]), exist_ok=True)
```
STMiki, line 298: `os.path.dirname`

STMiki (approved):
Don't mind me, this is just me giving too much attention to minor detail... :upside_down_face:

generators/protocol_block_ids.py
```diff
@@ -0,0 +1,303 @@
...
+    parser.add_argument("-f", "--format", help="format the code", action=argparse.BooleanOptionalAction)
+    args = parser.parse_args()
+    if args.input != "blocks.json":
+        print("I hope you know what you are doing, cause for now the script only works with the blocks.json file")
+    options["input"] = args.input
+    options["output"] = args.output
+    formated = args.format
+    return options
+
+# this is the main function
```
huntears, line 293: Thank you captain obvious xd

huntears (changes requested):
Where is the generated code ?

STMiki (approved):
👌

generators/protocol_block_ids.py
Trompettesib: You're welcome ^^

Trompettesib:
I almost did some tests but then I realize that I have already passed more than twice the time allowed for this task

Trompettesib:
What append if the file blocks.json does not exist ?

STMiki (changes requested):

cubic-server/Server.cpp
```diff
@@ -145,3 +160,42 @@ void Server::_stop()
         delete command;
     LINFO("Server stopped");
 }
+
+void Server::_downloadFile(const std::string &url, const std::string &path)
+{
+    if (std::filesystem::exists(path)) {
+        LDEBUG("File " << path << " already exists. Skipping download");
```
STMiki, line 167:
```suggestion
        LDEBUG("File " << path << " already exists. Skipping download");
        return;
```
and then remove the else 💢

cubic-server/protocol_id_converter/blockStates.cpp
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+        b.baseProtocolId = block.value()["states"][0]["id"];
+        b.maxProtocolId = block.value()["states"][block.value()["states"].size() - 1]["id"];
+        this->_blocks.push_back(b);
+    }
+}
+
+Blocks::BlockId Blocks::GlobalPalette::fromBlockToProtocolId(Blocks::Block block) const {
+    auto internalBlock = std::find_if(this->_blocks.begin(), this->_blocks.end(), [&block](Blocks::InternalBlock b) {
```
STMiki, line 38:
```suggestion
    auto internalBlock = std::find_if(this->_blocks.begin(), this->_blocks.end(), [&block](const Blocks::InternalBlock &b) {
```

cubic-server/protocol_id_converter/blockStates.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (block.properties.size() != internalBlock->properties.size()) {
+        LERROR("Block properties size mismatch with block " << block.name);
+        return 0;
+    }
+    if (block.properties.size() == 0 && internalBlock->properties.size() == 0)
+        return internalBlock->baseProtocolId;
+
+    BlockId id = internalBlock->baseProtocolId;
+    for (auto property : block.properties) {
+        auto internalProperty = std::find_if(internalBlock->properties.begin(), internalBlock->properties.end(), [&property](Blocks::InternalProperty p) {
```
STMiki, line 54:
```suggestion
        auto internalProperty = std::find_if(internalBlock->properties.begin(), internalBlock->properties.end(), [&property](const Blocks::InternalProperty &p) {
```

cubic-server/protocol_id_converter/blockStates.cpp
```diff
@@ -0,0 +1,88 @@
...
+            return 0;
+        }
+        id += internalProperty->baseWeight * (value - internalProperty->values.begin());
+    }
+    return id;
+}
+
+Blocks::Block Blocks::GlobalPalette::fromProtocolIdToBlock(Blocks::BlockId id) const {
+    for (auto b : this->_blocks) {
+        if (id >= b.baseProtocolId && id <= b.maxProtocolId) {
```
STMiki, line 73:
```suggestion
        if (id < b.baseProtocolId || id > b.maxProtocolId)
                continue;
```

cubic-server/protocol_id_converter/blockStates.hpp
```diff
@@ -0,0 +1,65 @@
...
+    class GlobalPalette {
+        public:
+            GlobalPalette() {};
+            ~GlobalPalette() {};
+            /**
+             * @brief Convert a block to a protocol id
+             * @param block The block to convert
+             * @return The protocol id of the block
+             */
+            BlockId fromBlockToProtocolId(Block block) const;
```
STMiki, line 50:
```suggestion
            BlockId fromBlockToProtocolId(const Block &block) const;
```

STMiki:
It fails the test, and you could have a base class with inerhitance from both items and block, I think, they are really similar

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM mcr.microsoft.com/devcontainers/cpp:0-ubuntu-22.04
```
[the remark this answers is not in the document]
Trompettesib: It's if you want to use github codespace. It's the default image.

.github/workflows/publish.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       if: ${{ !contains(github.repository, 'self-hosted') }}
       run: |
         sudo apt-get update
-        sudo apt-get install -y gcc-12 g++-12 cmake pkg-config gtkmm-3.0
+        sudo apt-get install -y gcc-12 g++-12 cmake pkg-config gtkmm-3.0 libcurl4-openssl-dev
```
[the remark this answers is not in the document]
Trompettesib: Here there isn't clang ^^

Trompettesib:
> Do you still need `generators/requirements.txt` ?

yes because I am using some of the libraries in the python file.
Maybe in the future, we will rethink about the code generation and reuse the python file

.devcontainer/Dockerfile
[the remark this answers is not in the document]
STMiki: I mean the file 😆

cubic-server/protocol_id_converter/blockStates.cpp
Trompettesib: Isn't this dangerous to let a return keyword in the for loop without having the if scope ?

huntears: Nop

cubic-server/Server.cpp
huntears: I don't think that's needed to be fair

.github/workflows/publish.yml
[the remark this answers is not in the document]
STMiki: I need more sleep...

cubic-server/protocol_id_converter/blockStates.hpp
```diff
@@ -0,0 +1,65 @@
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include "nlohmann/json.hpp"
+
+namespace Blocks {
+    // Using int32_t because it's the same size as a varint
+    typedef int32_t BlockId;
```
STMiki, line 9: Either use this everywhere and remove `GlobalBlockId` in `types.hpp` or use `GlobalBlockId` here

Trompettesib: but GlobalBlockId is a uint16_t...
```cpp
typedef uint16_t GlobalBlockId;
```

Trompettesib: How do I need to change it ?

STMiki: Wut ?

STMiki: wtf, I use this in varInt, this should be int32_t wtf

Trompettesib: https://github.com/CubicMC/cubic-server/blob/e92ff032f8bb4aa6d99d9cb207a99e04d5ba5d89/cubic-server/types.hpp#L15

Trompettesib: Do you want me to change that ?

STMiki: Ok no, I add them to the palette and convert them to an `int32_t` so you can just transform them directly

huntears (approved):

STMiki (approved):
👌

PR #109, opened 2023-02-11
Reorder functions & structures

NekoSelena:


Trompettesib (approved):
👌

huntears (commented):
This is fine, but this branch is based of the 1.19.3 branch which is still not merged, so let's wait for it to be merged before merging this one.

Trompettesib (approved):
blood everywhere on me
merge was so painfull, I jumped out of the window 4 times

cubic-server/protocol/ClientPackets.hpp
```diff
@@ -84,9 +77,102 @@ namespace protocol
...
+        u128 player_uuid;
+        double x;
+        double y;
+        double z;
+        uint8_t yaw;
+        uint8_t pitch;
+    };
+    std::shared_ptr<std::vector<uint8_t>> createSpawnPlayer(const SpawnPlayer &);
+
+    enum class EntityAnimationID : int32_t {
```
STMiki, line 106: use `uint8_t` instead of `int32_t`

STMiki (changes requested):
👌

cubic-server/protocol/ClientPackets.cpp
```diff
@@ -20,30 +56,143 @@ std::shared_ptr<std::vector<uint8_t>> protocol::createPingResponse(const PingRes
...
 
-std::shared_ptr<std::vector<uint8_t>> protocol::createLoginSuccess(const LoginSuccess &in)
+std::shared_ptr<std::vector<uint8_t>> protocol::createEntityAnimation(EntityAnimationID animId, int32_t entityID)
 {
     std::vector<uint8_t> payload;
-    serialize(payload, in.uuid, addUUID,
-        in.username, addString,
-        in.numberOfProperties, addVarInt
+    serialize(payload, entityID, addVarInt,
+        (uint8_t)animId, addByte);
```
STMiki, line 83: remove cast
```suggestion
        animId, addByte);
```

STMiki (approved):
👌

Trompettesib:
we need to find a way for using correctly enums without having to cast

Trompettesib (approved):
👌

cubic-server/protocol/ClientPackets.cpp
Trompettesib: We can't

Trompettesib: we need to cast somewhere

cubic-server/protocol/ClientPackets.cpp
```diff
@@ -79,8 +79,10 @@ std::shared_ptr<std::vector<uint8_t>> protocol::createSpawnPlayer(const SpawnPla
 std::shared_ptr<std::vector<uint8_t>> protocol::createEntityAnimation(EntityAnimationID animId, int32_t entityID)
 {
     std::vector<uint8_t> payload;
-    serialize(payload, entityID, addVarInt,
-        animId, addByte);
+    serialize(payload,
+        entityID, addVarInt,
+        (uint8_t)animId, addByte
```
STMiki, line 84: Same `uint8_t` 🤡

STMiki (changes requested):

cubic-server/protocol/ClientPackets.cpp
Trompettesib: But I can't remove the cast. If I do so, because animId is a EntityAnimationID, and not a uint8_t it will not compile

Trompettesib: We have the same problem everywhere else

STMiki: Oh yeah, enum class things 😅

STMiki (approved):

PR #110, opened 2023-02-12
Complex World generation based on different event

STMiki:
Big change, did multiple task in this one.

Bug fixes
- Fixed chunk serialization (biome too)
- Fix connection when the server is starting
- Clean ctrl+c / stop
- Added a mutex in Dimension for the Chunk map
- Changed pretty much every thread holder to the correct place (ex: WorldGroup thread was owned by the server, changed to WorldGroup)

Features:
- Close dialog on server close, with a good call to close the server
- Bumpy generation with noise map and the seed, and some decoration
- Multi threaded generation
- Sending chunks when the player move

Debug:
- Added a dump chunk command

<hr />

OFFICIAL GENERATION STEPS:
- [x] Terrain shaping: Which blocks are solid
- [x] Water filling: Determine and fill river, ocean or lake
- [x] Surface layer: top blocks with grass, dirt, sand, etc...
    - [ ] depending on the biome
- [ ] Features and structures: Tree, ores, villages

Haha simplified generation go brrr


CMakeLists.txt
```diff
@@ -48,9 +55,10 @@ pkg_check_modules(GTKMM REQUIRED gtkmm-3.0)
 add_executable(${CMAKE_PROJECT_NAME})
 add_subdirectory(cubic-server)
 
-set(CMAKE_EXE_LINKER_FLAGS "-static-libgcc -static-libstdc++")
```
huntears, line 51: Why did you remove the static linking ?

STMiki: Oops 😅

cubic-server/Player.hpp
```diff
@@ -119,6 +145,7 @@ class Player : public Entity
     uint8_t _gamemode;
     TickClock _keepAliveClock;
     bool _isFlying;
+    std::unordered_map<Position2D, ChunkStateHolder> _chunks;
```
Trompettesib, line 148: each player has a map of rendered or sended chunks
each dimension has a map of loading chunk

Trompettesib (changes requested):

Trompettesib (commented):
Things will change so I will review after

Trompettesib (approved):
👌 😅

huntears (approved):
augh

PR #112, opened 2023-02-21
Op command

PierrickMarguin:
/op

### DIFF
--- a/cubic-server/command_parser/commands/Op.cpp
+++ b/cubic-server/command_parser/commands/Op.cpp
@@ -1,0 +1,30 @@
+#include "Op.hpp"
+#include "Server.hpp"
+
+using namespace command_parser;
+
+void Op::autocomplete(std::vector<std::string>& args) const {
+    logging::Logger::get_instance()->info("autocomplete op");
+}
+
+void Op::execute(std::vector<std::string>& args) const {
+    if (args.empty()) {
+        logging::Logger::get_instance()->info("Need a player name");
+    } else {
+        if (args.size() != 1)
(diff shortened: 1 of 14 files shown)


Trompettesib (dismissed):
👌

cubic-server/command_parser/commands/Op.cpp
```diff
@@ -0,0 +1,34 @@
...
+using namespace command_parser;
+
+void Op::autocomplete(std::vector<std::string>& args, Player *invoker) const {
+}
+
+void Op::execute(std::vector<std::string>& args, Player *invoker) const {
+    if (invoker && !invoker->isOperator())
+        return;
+    if (args.empty()) {
+        LINFO("Need a player name");
```
STMiki, line 13: Add a `return` and remove the `else`

cubic-server/command_parser/commands/Op.cpp
```diff
@@ -0,0 +1,34 @@
...
+    } else {
+        if (args.size() != 1)
```
STMiki, line 15: Same

STMiki: Is the goal to reproduce the exact same behavior as the op command ? If not, you could support multiple op at once `/op [username1] [usernameX...]`

PierrickMarguin: I reproduced the same behavior of the op command of the official server so it cannot give operator privileges to multiple players.

STMiki (approved):

PR #113, opened 2023-02-22
chat command handling

Trompettesib:


Trompettesib:
I need to comment what I did btw

cubic-server/command_parser/commands/QuestionMark.hpp
```diff
@@ -5,11 +5,11 @@
 
 namespace command_parser {
 struct QuestionMark : public CommandBase {
-    QuestionMark() : CommandBase("?", "/help [<command>] (question mark)") {}
+    QuestionMark() : CommandBase("?", "/help [<command>] (question mark)", 0) {}
```
STMiki, line 8: /? instead of /help

cubic-server/protocol/ServerPackets.hpp
```diff
@@ -26,6 +26,7 @@ namespace protocol
         QueryBlockEntityTag = 0x01,
         ChangeDifficulty = 0x02,
         ChatMessage = 0x03,
+        ChatCommand = 0x04,
```
STMiki, line 29: 0x03, we are still in 1.19 (I think)
And yes, ChatMessage should be 0x04, but that does not fix the chat...

STMiki (commented):
Why did you implement the permissions level system ? didn't we discuss that it will be reworked ? With like real permissions and stuff

cubic-server/protocol/Structures.hpp
```diff
@@ -50,5 +50,11 @@ namespace protocol
         int32_t type;
         nbt::Compound data;
     };
+
+    struct argumentSignature
```
huntears, line 54: That should be in PascalCase

huntears (changes requested):

huntears (approved):

STMiki (approved):

PR #115, opened 2023-03-02
Improved logging macro and cleaned existing call to said macro

### LINKED ISSUE #96: Add string stream to logging to be able to use << operator with various types

STMiki:
You can call the logger like this:
```cpp
LINFO("I am", " dumb"); // Multiple arguments up to 9
LINFO("I am" << "dumb"); // chain << operator, up to ...
LINFO("I ", "am" << " ", "dumb"); // You can use both at the same time

LINFO(); // Output: "File.cpp:42 Log called without any arguments !"
```
Fix #96

### DIFF
--- a/cubic-server/Chat.cpp
+++ b/cubic-server/Chat.cpp
@@ -7,7 +7,6 @@
 
 Chat::Chat()
 {
-    _log = logging::Logger::get_instance();
 }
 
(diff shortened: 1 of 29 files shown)


Trompettesib (approved):
Seems perfect to me 👌

PR #119, opened 2023-03-08
Make the code base use the new generated block palette

huntears:


cubic-server/world_storage/ChunkColumn.hpp
```diff
@@ -83,7 +83,6 @@ class ChunkColumn
     ChunkColumn(const Position2D &chunkPos);
     ~ChunkColumn();
 
-    void updateBlock(Position pos, Block block);
```
Trompettesib, line 86: not sure about that

Trompettesib (approved):
Looks good to me

cubic-server/world_storage/ChunkColumn.hpp
STMiki: Do we still use the Block class ?

huntears:
> Good, but shouldn't the blocks folder be generated by cmake ?

I prefer doing by hand, just to check that everything is alright, and since it only changes when there's a version change I would say it is fine.

PR #121, opened 2023-03-14
Fix whitelist behavior

NekoSelena:
- [x] update the configuration, so that the `enforce-whitelist` field has the correct behavior (“enforce” the whitelist on the server)
- [x] add the `whitelist` field to the configuration, allowing to activate or not the whitelist
- [x] update the functions managing the whitelist to rectify the behaviors of the modified fields of the configuration
  - [x] enforceWhitelistOnReload()
  - [x] whitelist behavior in onLoginStart (Client connexion handling)

### DIFF
--- a/cubic-server/Server.cpp
+++ b/cubic-server/Server.cpp
@@ -105,6 +105,20 @@ void Server::stop()
     this->_running = false;
 }
 
+void Server::forEachWorldGroup(std::function<void(WorldGroup &)> callback)
+{
+    for (auto &[_, worldGroup] : this->_worldGroups)
+        callback(*worldGroup);
+}
+
+void Server::forEachWorldGroupIf(std::function<void(WorldGroup &)> callback, std::function<bool(const WorldGroup &)> predicate)
+{
+    for (auto &[_, worldGroup] : this->_worldGroups) {
+        if (predicate(*worldGroup))
+            callback(*worldGroup);
+    }
+}
+
 void Server::_acceptLoop()
 {
     struct pollfd poll_set[1];
--- a/cubic-server/Server.cpp
+++ b/cubic-server/Server.cpp
@@ -38,12 +38,12 @@ Server::Server():
     _running(false)
 {
     _config.parse("./config.yml");
-    _whitelistData = _whitelistHandler.parseWhitelist("./whitelist.json");
+    _whitelist = WhitelistHandling::Whitelist();
     _host = _config.getIP();
     _port = _config.getPort();
     _maxPlayer = _config.getMaxPlayers();
     _motd = _config.getMotd();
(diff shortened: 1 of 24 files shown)


Trompettesib (dismissed):
Things in whitelist have no sens but all the code is very good !

Trompettesib:
the fact that the object whitelist does not store the json into it is quite strange

cubic-server/Server.cpp
```diff
@@ -208,3 +210,20 @@ void Server::_downloadFile(const std::string &url, const std::string &path)
...
+/*
+** If the server gets a /reload, players not on the whitelist
+** must be kicked from the server if enforce-whitelist is true
+** & the whitelist is in effect
+*/
+void Server::enforceWhitelistOnReload() {
+    if (_whitelist && _enforceWhitelist) {
+        for (auto &client : _clients) {
+            if (!_whitelistHandler.isPlayer(client->getPlayer()->getUuid(), client->getPlayer()->getUsername(), _whitelistData).first) {
+                client->stop("You are not whitelisted on this server.");
```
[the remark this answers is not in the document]
STMiki: It seems I was drunk when I created the stop functions, don't use them, most of them are useless.

Will fix later :sweat:

cubic-server/Server.hpp
```diff
@@ -45,10 +46,20 @@ class Server
 
     void stop();
 
+    void enforceWhitelistOnReload();
+
     const Configuration::ConfigHandler &getConfig() const {
         return _config;
     }
 
+    WhitelistHandling::Whitelist getWhitelist() const {
```
Trompettesib, line 55: This is strange that we can't use a const ref here but it's ok

Trompettesib (approved):
👌

cubic-server/Server.hpp
NekoSelena: You know it ends up crashing if we make it const.

STMiki: It's because `isPlayerWhitelisted` is not const

NekoSelena: Yes, because it can't be because of the iterator, it crashes if it is const

STMiki: And thats why you use auto on iterator.

Try with `const_iterator` 😉

STMiki: Or just use auto

NekoSelena: Fixed 👍

Trompettesib (approved):
👌

STMiki (approved):

huntears (approved):

Trompettesib (approved):

STMiki (approved):

PR #125, opened 2023-03-24
Clean send functions in player and client object

Trompettesib:


Trompettesib:
I think the way we are using SynchronizePosition packet is not good but it's just a gess

cubic-server/Player.cpp
```diff
@@ -1015,3 +880,175 @@ void Player::_processKeepAlive()
...
+        this->_dim->removePlayerFromLoadingChunk({x, z}, this);
+        this->_chunks.erase({x, z});
+        return;
+    }
+
+    this->sendUnloadChunk(x, z);
+    this->_chunks.erase({x, z});
+}
+
+void Player::_synchronizePostion(Vector3<double> pos)
```
huntears, line 1043: `Postion` :)

huntears: Also I would prefer if that function was called more like `teleport` or something like that.
Synchronize position is more just the protocol.

Trompettesib: but teleport is another paquet, this is why I said that I think we are using synchronizePosition badly
[teleport_entity](https://wiki.vg/index.php?title=Protocol&oldid=18067#Teleport_Entity)
[synchronize_position](https://wiki.vg/index.php?title=Protocol&oldid=18067#Synchronize_Player_Position)

huntears: This is not a function to send a packet, that is a function to move the player somewhere else, therefore a method called teleport in the player would be appropriate in this case.

huntears (approved):

STMiki (approved):

STMiki (approved):

PR #128, opened 2023-03-31
Loot tables

PierrickMarguin:


cubic-server/loot_tables/rolls/Roll.hpp
```diff
@@ -2,56 +2,56 @@
...
+/*
+  Roll result is the result of a roll
+  it contains the number of rolls to be done and a probability(0=none;1=always) for each roll to be added
+*/
+struct RollResult {
+    RollResult(int64_t nbr, double probability);
+    ~RollResult() = default;
+
+    int64_t _nbr;
+    double _probability;
```
STMiki, line 24: This is public by default, so no leading _

STMiki (changes requested):
No ; after a namespace closing bracket

STMiki (changes requested):

cubic-server/loot_tables/LootTable.cpp
```diff
@@ -0,0 +1,91 @@
...
+        if (!pool->isValid()) {
+            this->_validity = false;
+            return;
+        }
+    }
+    this->_validity = true;
+}
+
+// calls every Pool poll function with context
+LootTablePoll LootTable::poll(LootContext *context) const
```
Trompettesib, line 80: why do you need a raw pointer here ?

cubic-server/loot_tables/conditions/Alternative.cpp
```diff
@@ -0,0 +1,20 @@
...
+namespace LootTable {
+namespace Condition {
+Alternative::Alternative(const nlohmann::json &condition)
+{
+    (void) condition;
+
+    this->setValidity(true);
+}
+
+bool Alternative::verify(const LootContext *context) const
```
Trompettesib, line 12: Same why do you need a raw pointer here ?

Trompettesib (commented):
some basic things to change but overall I think that the code quality is good

Trompettesib:
Maybe you want to put some variable in private or in protected in some classes (the _variables) ?

cubic-server/loot_tables/LootTable.cpp
PierrickMarguin: It is best to provide a LootContext when applying a loot table but it is not mandatory so we could pass a nullptr

huntears (dismissed):

Trompettesib (approved):

STMiki (dismissed):

PR #142, opened 2023-04-24
Add SpawnEntity packet

NekoSelena:
This packet create entities client-side

### DIFF
--- a/cubic-server/protocol/ClientPackets.hpp
+++ b/cubic-server/protocol/ClientPackets.hpp
@@ -242,7 +242,7 @@ struct SpawnEntity {
     int16_t velocityY;
     int16_t velocityZ;
 };
-std::shared_ptr<std::vector<uint8_t>> createSpawnEntity(const SpawnEntity &);
+std::unique_ptr<std::vector<uint8_t>> createSpawnEntity(const SpawnEntity &);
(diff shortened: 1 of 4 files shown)


STMiki (dismissed):

cubic-server/protocol/ClientPackets.hpp
```diff
@@ -101,6 +102,148 @@ struct PingResponse {
 };
 std::shared_ptr<std::vector<uint8_t>> createPingResponse(const PingResponse &);
 
+struct SpawnEntity {
+    int32_t entityId;
+    u128 entityUuid;
+    enum class EntityType : int32_t {
```
huntears, line 108: Are those IDs correct ? Because there is the Camel which is not in 1.19.3 taking an ID, which sounds weird.

STMiki: it is, just locked behind the 1.20 experimental update flag.

Trompettesib (dismissed):
Later, we will use the registries.json to get each entities id's but for now and the beta it's good

huntears (dismissed):

STMiki (approved):

huntears (approved):

PR #150, opened 2023-05-05
Fix FreeBSD compilation

huntears:
It currently compiles but it segfaults instantly once a player joins.


huntears:
Closing as #157 is based on this branch

PR #158, opened 2023-05-11
Meson support

Matthieu-Sauer:
This PR should enable meson compilation support. Meson has by default more warnings enabled, it might be useful to track them.
Improvements over cmake:
Submodules! meson uses submodules to delegate pulling from a repo to other scripts. this enables better maintenability of the submodules
Dependency detection!
Meson can choose if we want to use local submodules or not. this enables the user to install dependencies on his machine instead of pulling an entire repository.
Better organisation!
Submodules have their own folder (it's mandatory), same goes for the build. the build is automatically added to a .gitignore if the meson is executed inside a repo git (no risk of pushing it by accident).

Things meson looses over cmake:
Wildcards! meson do not support them
Script-like behaviour! Meson is not turing complete, meaning it cannot be considered a scripting language (like cmake). you cannot define custom functions.

There is still the need to modify the readme (to add meson compilation in it), and add the test files to the compilation process.


subprojects/perlin_noise.wrap
```diff
@@ -0,0 +1,7 @@
+[wrap-git]
+url = https://github.com/Matthieu-Sauer/PerlinNoise.git
+revision = master
```
huntears, line 3: Please specify a specific version or commit here.
It would be bad if behaviour changed by itself.

subprojects/rapidcheck.wrap
```diff
@@ -0,0 +1,7 @@
+[wrap-git]
+url = https://github.com/emil-e/rapidcheck.git
+revision = master
```
huntears, line 3: Same here, please provide a specific version

huntears (commented):
That sounds pretty good, I'll see with the rest of the core team to see what to do of your PR :)

huntears:
I also realised that you only allow testing if in debug, but I think being able to build tests in release would also be really good.
You might also need a way to enable LTOs if possible when in release.

meson.build
```diff
@@ -0,0 +1,117 @@
...
+yaml     = dependency('yaml-cpp',      required: false)
+json     = dependency('nlohmann_json', required: false)
+noise    = dependency('perlin_noise',  required: false)
+argparse = dependency('argparse',      required: false)
+crcpp    = dependency('CRCpp',         required: false)
+
+if get_option('buildtype') == 'debug' and get_option('test')
+  gtest = dependency('gtest', main: true, required: false)
+  rapidcheck = dependency('rapidcheck', required: false)
+endif
```
eli-schwartz, line 56: Since testing is already a build option, why not just always create the test target when tests are requested?

meson.build
```diff
@@ -0,0 +1,117 @@
...
+cmake = import('cmake')
+
+if not yaml.found()
+  yaml_subproj = cmake.subproject('yaml-cpp')
+  yaml = yaml_subproj.dependency('yaml-cpp')
+endif
+
+if not json.found()
+  json = dependency('nlohmann_json')
+endif
```
eli-schwartz, line 71: Why do you first check for it with required: false, then check for it again with required: true if it is not found the first time? Just remove "required: false" from the first lookup. :)

meson.build
```diff
@@ -0,0 +1,117 @@
...
+
+# we need to create a patch fot this one.
+if not noise.found()
+  noise = dependency('perlin_noise')
+endif
+
+if not argparse.found()
+  argparse = dependency('argparse')
+endif
```
eli-schwartz, line 80: Here too.

meson_options.txt
```diff
@@ -0,0 +1,7 @@
+#===============
+# custom options
+#===============
+
+option('libcpp_clang',     type : 'boolean', value : false, description : 'use libc++ if compiling with clang and the library is available')
+option('test',             type : 'boolean', value : false, description : 'activate tests (debug only)')
+option('download_missing', type : 'combo',  choices: ['minimal', 'full'], value : 'minimal', description : 'activate tests (debug only)')
```
eli-schwartz, line 7: What does this option do? Its description is based on the description of the "test" option...

(Also: github is complaining about the newline ending of the file.)

subprojects/perlin_noise.wrap
Matthieu-Sauer: Change will be implemented :)

meson.build
Matthieu-Sauer: Indeed, I did not know this was possible at the time, I will modify this accordingly

meson.build
Matthieu-Sauer: At first, it was so I could choose between using system deps and wrap system. I will do this for the heaviest dependencies (boost i'm looking at you), with the default being recompiling boost if needed, otherwise require a system installation. I don't know if that could be considered good practice?

meson_options.txt
Matthieu-Sauer: actually nothing, it was supposed to be used to activate tests, however since there is already an option to do so, it will be removed.
download_missing will be used (and renamed) to activate system-only usage of heavy libraries.

meson.build
eli-schwartz: How does this let you choose between the two? Note that the default is already to try finding a system dependency first, and if permitted, fall back to a wrap dependency (by default if the dependency is required it can fallback, and cannot fallback if it's optional; this can be controlled with the `allow_fallback` kwarg).

If users want to enforce either choice, it's possible to do so by configuring the build with the option `--wrap-mode=` and the possible values:
- nofallback -- never use wraps
- forcefallback -- never use system dependencies 
- default
- ...

You can also force fallback for exactly a specific set of dependencies with e.g. `--force-fallback-for=boost, perlin_noise`.

eli-schwartz: You can at least limit the `required: false` to the cases where you need to specially handle it with cmake. :)

Matthieu-Sauer: indeed! that's what i'll do at first :) I will still try to create patches on the side as I feel it's good practice

eli-schwartz: Also yes, I'm very happy we (meson) were able to design a system that puts users and builders in control of how vendored vs. system dependencies works... It's something that has consistently bothered me in every single build system other than meson, and I've used a lot of them (not just autotools and cmake).

Matthieu-Sauer: true, I have only two things that bothers me with it: not enough dependency are using meson (so more cmake wrapper haha)
and the fact you need to solve each dependency one at a time if they are missing, it does not tell "hey, I need those 4 dependencies but you don't have 3 of them, come back".
Since I am trying to port this project from cmake to meson, I clearly see the difference in readability, and in pushed good practice. I hope it will not become a cluster of legacy things as cmake did :/
autotools is just (sadly) the worst I tried, too slow, checking for useless things...
scons... well scons.
I really like this way of managing dependencies tho, just saying "find it, if you don't, it's there" making the use of another program for dependency management useless is perfect (no more conan, nuget...)

huntears:
We are done with the current version on this project and are currently working on a complete rewrite, closing this.

PR #161, opened 2023-05-24
Removing gtkmm from required dependency

STMiki:


Trompettesib (approved):

huntears:
It was my work but ok it just works ( ‾ʖ̫‾)

PR #169, opened 2023-05-31
Fix test in Release mode and control what boost download/compile

STMiki:
- Test only compile if GTEST is enabled
- Restrict what library boost downloads/compile

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,9 +52,11 @@ if (CMAKE_BUILD_TYPE MATCHES Release)
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


huntears:
Wouldn't it be easier to get a big archive of the boost version we need and just query it when needed?
Because even with only what we need this takes ages to fetch all repos.

STMiki:
> Wouldn't it be easier to get a big archive of the boost version we need and just query it when needed? Because even with only what we need this takes ages to fetch all repos.

We don't really care, because this will be a one time download if boost is not already installed

Trompettesib (approved):

huntears (approved):

PR #173, opened 2023-06-02
Various fixies

Trompettesib:


cubic-server/loot_tables/CMakeLists.txt
```diff
@@ -13,27 +13,8 @@ download_and_check_file(
...
-    COMMAND mv -n ${CMAKE_BINARY_DIR}/loot_tables ${CMAKE_BINARY_DIR}/assets/loot_tables
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-)
-
-execute_process(
-    COMMAND rm -r ${CMAKE_BINARY_DIR}/loot_tables
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-)
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/assets)
+file(ARCHIVE_EXTRACT INPUT ${CMAKE_BINARY_DIR}/loot_tables.tar.gz DESTINATION ${CMAKE_BINARY_DIR}/assets)
```
STMiki, line 17: Wasn't this already fixed ?

STMiki (approved):

STMiki (dismissed):

cubic-server/Chat.hpp
```diff
@@ -21,7 +21,7 @@ class Chat {
      * @param message
      * @param from
      */
-    void sendPlayerMessage(const chat::Message &message, const Player &from);
+    void sendPlayerMessage(const chat::Message &message, const std::shared_ptr<Player> from);
```
huntears, line 24: Why did you change everything reference to the player to shared_ptrs?

cubic-server/Dimension.cpp
```diff
@@ -159,7 +159,7 @@ void Dimension::loadOrGenerateChunk(int x, int z, std::shared_ptr<Player> player
                 const Vector3<double> chunkPos = {(double) x * 16, pos.y, (double) z * 16};
                 current_min = std::min(current_min, pos.distance(chunkPos));
             }
-            return static_cast<size_t>(std::ceil(current_min));
+            return static_cast<int>(std::ceil(current_min));
```
huntears, line 162: Why change it to an int? Is there a reason? Because `current_min` cannot be negative in the first place.

cubic-server/Entity.cpp
```diff
@@ -96,26 +99,19 @@ void Entity::setRotation(const Vector2<uint8_t> &rot)
...
-std::shared_ptr<World> Entity::getWorld() const { return _dim->getWorld(); }
+Vector3<double> &Entity::getPosition() { return _pos; }
+Vector2<uint8_t> &Entity::getRotation() { return _rot; }
+Vector3<double> &Entity::getLastPosition() { return _lastPos; }
+Vector2<uint8_t> &Entity::getLastRotation() { return _lastRot; }
 
 std::shared_ptr<WorldGroup> Entity::getWorldGroup() const { return _dim->getWorld()->getWorldGroup(); }
-
+std::shared_ptr<World> Entity::getWorld() const { return _dim->getWorld(); }
+std::shared_ptr<Dimension> Entity::getDimension() const { return _dim; }
```
huntears, line 109: Please make sure to return constant `shared_ptr`s when in a constant method.

cubic-server/EntityType.hpp
```diff
@@ -0,0 +1,381 @@
...
+        return "minecraft:zombie_villager";
+    case EntityType::ZombifiedPiglin:
+        return "minecraft:zombified_piglin";
+    case EntityType::Player:
+        return "minecraft:player";
+    case EntityType::FishingBobber:
+        return "minecraft:fishing_bobber";
+    default:
+        return "minecraft:unknown";
+        break;
```
huntears, line 377: That break is useless.

cubic-server/EntityType.hpp
```diff
@@ -0,0 +1,381 @@
...
+    ZombieHorse = 114,
+    ZombieVillager = 115,
+    ZombifiedPiglin = 116,
+    Player = 117, // SpawnPlayer
+    FishingBobber = 118,
+};
+
+inline std::string getEntityTypeName(EntityType type)
+{
+    switch (type) {
```
huntears, line 136: Why not just have an `std::array` and then just get the name through a jump into it?

cubic-server/Player.cpp
```diff
@@ -56,9 +55,7 @@ Player::Player(std::weak_ptr<Client> cli, std::shared_ptr<Dimension> dim, u128 u
     _keepAliveClock.start();
     _heldItem = 0;
 
-    this->_uuidString = this->getUuid().toString();
```
huntears, line 59: Please don't remove that assignment, it is really usefull as player UUIDs are used as strings frequently and that would avoid transforming it to an `std::string` everytime.

cubic-server/command_parser/CommandParser.cpp
```diff
@@ -7,11 +7,14 @@
 #include "World.hpp"
 #include "logging/logging.hpp"
 
-void command_parser::parseCommand(std::string &command, Player *invoker)
+void command_parser::parseCommand(std::string &command, std::shared_ptr<Player> invoker)
```
huntears, line 10: Why do you change everything to a shared_ptr for the commands? The reference counting while work for every single function call with that, and I am pretty sure that the commands don't need to own a Player, but only to reference to it, so a reference with an std::optional or a raw pointer is fine here.

cubic-server/thread_pool/PriorityThreadPool.hpp
```diff
@@ -114,7 +114,7 @@ class PriorityThreadPool {
         {
             const std::lock_guard<std::mutex> _guard(_toolBox.queueProtection);
             jobId = ++_toolBox.totalJobsPushed;
-            _toolBox.jobQueue.emplace_back(jobId, priority, std::move(realJobList));
+            _toolBox.jobQueue.emplace_back(PriorityThreadPoolUtility::Job {jobId, priority, std::move(realJobList)});
```
huntears, line 117: Why create an object here for emplace? It was working fine without, and it would construct the object inplace in the jobqueue, which is faster.

cubic-server/types.cpp
```diff
@@ -35,3 +36,11 @@ u128 u128::fromShortString(const std::string &str)
     std::string second = std::string(it, str.end());
     return u128({std::stoull(first, 0, 16), std::stoull(second, 0, 16)});
 }
+
+u128 u128::random()
+{
+    std::random_device rd;
+    std::mt19937_64 gen(rd());
```
huntears, line 43: Please create the `std::random_device` and the `std::mt19937_64` for caching, it is unnecessary to create them every single time and defeats the purpose of randomness.

huntears (changes requested):
Why so many `std::shared_ptr`?

cubic-server/Chat.hpp
STMiki: I remember getting angry over this, but I don't remember why...

cubic-server/Dimension.cpp
STMiki: It's not a problem with current_min, but the threadpool job priority function which take an int, and it may throw a warning because of the narrowing.

cubic-server/thread_pool/PriorityThreadPool.hpp
STMiki: I think it was a clang error ? But it was so long ago that this is a total guess🤡

huntears: I'll check today

STMiki: I was right, this generate an error with clang, but only because the struct was declared in a weird way...

cubic-server/EntityType.hpp
STMiki: I don't think this is important, this function is only called when there is an hover event on a chat message, and it's already decently fast.

cubic-server/Chat.hpp
STMiki: I can't remember, I am reversing this

cubic-server/Player.cpp
STMiki: You can always redo this, but as of right now this is not used anywhere.

huntears (approved):

PR #175, opened 2023-06-13
Fix the player position synchronization between client and server 

Thomas-Harit:
…rue on initialize

### DIFF
--- a/cubic-server/Entity.hpp
+++ b/cubic-server/Entity.hpp
@@ -73,7 +73,6 @@ public:
     NODISCARD virtual const Vector3<double> &getPosition() const;
     NODISCARD virtual Vector2<uint8_t> &getRotation();
     NODISCARD virtual const Vector2<uint8_t> &getRotation() const;
-    NODISCARD virtual Vector2<float> getRotationDegree();
     NODISCARD virtual Vector2<float> getRotationDegree() const;
     NODISCARD virtual Vector3<double> &getLastPosition();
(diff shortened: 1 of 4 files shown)


STMiki (dismissed):

cubic-server/Entity.hpp
```diff
@@ -72,6 +73,8 @@ class Entity : public utility::SharedFromThis<Entity> {
     NODISCARD virtual const Vector3<double> &getPosition() const;
     NODISCARD virtual Vector2<uint8_t> &getRotation();
     NODISCARD virtual const Vector2<uint8_t> &getRotation() const;
+    NODISCARD virtual Vector2<float> getRotationDegree();
+    NODISCARD virtual const Vector2<float> getRotationDegree() const;
```
huntears, line 77: Please return a reference for those two functions

huntears (changes requested):
Really good overhaul, just one small thing to fix.

cubic-server/Entity.hpp
Thomas-Harit: Those functions return a locally declared variable, can't reference that :pepe_sad:

huntears: Then just create only one function like this:
```suggestion
    NODISCARD virtual Vector2<float> getRotationDegree() const;
```

STMiki (approved):

huntears (approved):

PR #177, opened 2023-06-19
Registry

STMiki:
Working registry created, and accessible from the `Server` class.

The nbt registry is only created once on startup and is reused when needed.

This mean that any change to the registry will throw / not work after initialization.


Trompettesib (approved):

huntears (changes requested):
Making getters and setters constexpr will just increase compile time, please mark them inline.
The rest of the PR is good, it would have been good to use the new nbt librarie it is fine for now :).

Trompettesib (approved):

huntears (approved):

Trompettesib (approved):

PR #180, opened 2023-07-11
Add the Birch tree

NekoSelena:
Adding the birch tree generation


NekoSelena:
Based on Pickup Item, do not take into account anything besides `birch.cpp` & `birch.hpp`

NekoSelena:
Closed because of messed up merges

PR #181, opened 2023-07-13
Add Jungle tree generation

NekoSelena:
The leaves algorithm is the same as the one for the birch tree (to be updated on birch)


Trompettesib (approved):
same as in birch pr

PR #184, opened 2023-07-19
Generic tree generation

NekoSelena:
Fixing #181 & #182 with a generic tree generation for the Oak, Birch & Jungle trees


Trompettesib (dismissed):
C BÔ

*It's beautiful but in french and badly written*

cubic-server/world_storage/ChunkColumn.hpp
```diff
@@ -105,6 +105,8 @@ class ChunkColumn {
     constexpr std::array<Section, NB_OF_SECTIONS> &getSections() { return _sections; }
     constexpr const std::array<Section, NB_OF_SECTIONS> &getSections() const { return _sections; }
 
+    const std::shared_ptr<Dimension> getDimension() const { return _dimension; }
```
STMiki, line 108:
```suggestion
    const std::shared_ptr<const Dimension> getDimension() const { return _dimension; }
```
`const std::shared_ptr<thing>` -> the pointer is const
`std::shared_ptr<const thing>` -> the object the pointer is pointing to is const

And shared_ptr has a const copy constructor from a non-const version, so this is the only thing you need to change

STMiki (changes requested):

huntears (dismissed):
There's just the thing that Miki mentionned, but the rest if fine

STMiki (approved):

huntears (approved):

PR #185, opened 2023-07-20
Fall damage

QuentindiMeo:
Implementation of fall damage for LivingEntity class, with bool method for Player; includes detection of soft blocks and immunity


huntears:
All the calculations are there but there is no call to them so no fall damages are applied ever.
Getting block softness should also not be a method of LivingEntity and it should use protocolIDs, not strings as this would hinder performance quite a bit.

PR #186, opened 2023-07-20
Redstone

QuentindiMeo:


cubic-server/RedstoneActivators.cpp
```diff
@@ -0,0 +1,49 @@
...
+        ;// TODO make activation noise (1273)
+    else // stone
+        ;// TODO make activation noise (1107)
+    // TODO when clock reaches _duration, launch deactivation
+}
+
+Button::Button(std::string material):
+    Lever(),
+    _material(material),
+    _duration(_material.find("stone") != std::string::npos ? 10 : 15) // ticks
```
STMiki, line 30: isn't that 20 ticks for a wooden button ?
And why use a string and not just an old plain bool ?

PR #192, opened 2023-08-03
Stop the server from playing a bat takeoff sound every 20 seconds

huntears:
It is so incredibly annoying >:(


STMiki (approved):
Ok lol

Trompettesib (approved):

PR #200, opened 2023-08-27
Add the feature "Forest Rock"

NekoSelena:
https://minecraft.fandom.com/wiki/Forest_rock

This generation is seed-based (size & position)


cubic-server/generation/generator.cpp
```diff
@@ -40,9 +35,22 @@ generation::Generator::GenerationNoise generation::Generator::getNoise(positionT
     // noise.noise2D.peaksAndValley = _noiseMaker.octave2D_11(_x, _z, octaves);
     noise.noise2D.weirdness = _noiseMaker.octave2D_11(_x, _z, octaves);
     noise.noise2D.trees = _noiseMaker.octave2D_11(x * 0.5, z * 0.5, 1);
+    noise.noise2D.rocks = _noiseMaker.octave2D_11(x * 0.6, z * 0.6, 1);
```
NekoSelena, line 38: Play with this line to change Rock generation density

huntears:
Looks like the generation is far from being good here ;w;

Also why do I need to go and comment and uncomment pieces of code to test this?

NekoSelena:
> Looks like the generation is far from being good here ;w;

The generation is supposed to be like this. Sometimes the rocks are not all round and perfect. Here is a bunch of test rocks I "placed" in the original game. As you can see, they are pretty much diverse and not round at all.
image
image
image

And this one in its original biome, from the original game generation :
image


> Also why do I need to go and comment and uncomment pieces of code to test this?

The code is commented for the people who want to test it, as the Forest rock only appears in the Old Taiga biomes, which are not implemented yet.

STMiki (dismissed):
mouais

Trompettesib (dismissed):

Trompettesib (approved):

STMiki (approved):

huntears:
Ok so you did a rebase but that invalidated all of the signed commits :/
Later we need to avoid rebases if commits are gpg signed as that would directly invalidate the signings.

huntears (approved):
It should be fine for the next delivery, but I really need to work on that feature library 💀

PR #205, opened 2023-09-08
Net compression

huntears:
I also made online-mode experimental, because the unstability would just crash directly with the compression.

### DIFF
--- a/cubic-server/CompressionUtils.cpp
+++ b/cubic-server/CompressionUtils.cpp
@@ -1,0 +1,91 @@
+#include <cstddef>
+#include <cstdint>
+#include <unistd.h>
+#include <vector>
+#include <zlib.h>
+
+#include "Server.hpp"
+
+static void add_buffer_to_vector(std::vector<uint8_t> &vector, const uint8_t *buffer, size_t length)
+{
+    for (size_t character_index = 0; character_index < length; character_index++) {
+        char current_character = buffer[character_index];
+        vector.push_back(current_character);
+    }
+}
+
+int compressVector(const std::vector<uint8_t> &source, std::vector<uint8_t> &destination)
+{
+    unsigned long source_length = source.size();
+    size_t destination_length = compressBound(source_length);
+
+    uint8_t *destination_data = (uint8_t *) malloc(destination_length);
+    if (destination_data == nullptr) {
+        return Z_MEM_ERROR;
+    }
+
+    Bytef *source_data = (Bytef *) source.data();
+    int return_value = compress2((Bytef *) destination_data, &destination_length, source_data, source_length, CONFIG["compression-level"].as<int>());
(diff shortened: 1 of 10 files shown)


huntears:
Just need to clean the code real quick and this should be ready to merge :)

huntears:
Ok so it looks like Krypton is incompatible with this implementation of the compression, I have no idea why though :/

Trompettesib (dismissed):
¯\\_(ツ)_/¯

cubic-server/CompressionUtils.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <vector>
+#include <zlib.h>
+
+#include "Server.hpp"
+
+static void add_buffer_to_vector(std::vector<uint8_t> &vector, const uint8_t *buffer, size_t length)
+{
+    for (size_t character_index = 0; character_index < length; character_index++) {
+        char current_character = buffer[character_index];
+        vector.push_back(current_character);
```
STMiki, line 13: ?
```suggestion
        vector.push_back(buffer[character_index]);
```

cubic-server/CompressionUtils.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+int compressVector(const std::vector<uint8_t> &source, std::vector<uint8_t> &destination)
+{
+    unsigned long source_length = source.size();
+    size_t destination_length = compressBound(source_length);
+
+    uint8_t *destination_data = (uint8_t *) malloc(destination_length);
```
STMiki, line 22: What about no ?

cubic-server/CompressionUtils.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return return_value;
+}
+
+bool decompressVector(const std::vector<uint8_t> &compressedBytes, std::vector<uint8_t> &uncompressedBytes, uint32_t size)
+{
+    unsigned full_length = size;
+    unsigned half_length = size / 2;
+
+    unsigned uncompLength = full_length;
+    char *uncomp = (char *) calloc(sizeof(char), uncompLength);
```
STMiki, line 40: Wsh

cubic-server/CompressionUtils.cpp
huntears: Yes, I will use that, cause zlib is a C library, I don't really have a choice :(

cubic-server/CompressionUtils.cpp
huntears: What is wrong with it?

huntears:
Fixed #208 , just a freaking of by one error :)))))))))

Trompettesib (approved):
works for me on my network and my computers

STMiki (approved):

PR #209, opened 2023-09-16
Foliage distance generation

Trompettesib:


huntears:
This just needs to be formatted properly and it should be fine.
Later we should look at handling distances when we place a leaf block, which we don't right now.

NekoSelena (approved):

huntears (approved):

QuentindiMeo (approved):
overall code is rather clean, looking good.
performance improvements can be made through avoiding branches and whatnot though

PR #212, opened 2023-09-17
Nbt in slot

STMiki:
nbt are in slot.

<s>**NOTE** the nbt parsing from the client is done, but sending a nbt from the server is not done, because serializing the nbt_tag_t is a pain and I will do it in the wrapper that I am working on.</s> NVM


cxx_flag_overrides.cmake
```diff
@@ -4,7 +4,7 @@ set(CMAKE_CXX_FLAGS_RELEASE_INIT "-O3 -DNDEBUG")
 set(CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT "-O2 -ggdb3 -fno-omit-frame-pointer -fsanitize=address -fsanitize=null")
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-array-bounds -Wno-stringop-overflow")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-array-bounds -Wno-stringop-overflow -pipe")
```
huntears, line 7: Did we do proper tests with this option though ? Because last time I thought it wasn't faster than without it.

huntears (commented):
Creating a macro for freeing the nbt tags could be a good idea, or we really need a C++ wrapper for that library

cxx_flag_overrides.cmake
STMiki: Nope, no question. just "add it it's faster" and I added it :laughing:

STMiki: Imagine if we could use shared_ptr :exploding_head:

cubic-server/Player.cpp
```diff
@@ -71,6 +72,20 @@ Player::Player(std::weak_ptr<Client> cli, std::shared_ptr<Dimension> dim, u128 u
...
+    constexpr std::string_view NAME = "[{\"text\":\"Cubic\",\"italic\":false}]";
+    constexpr std::string_view NAME_TAG = "Name";
+    constexpr std::string_view DISPLAY_TAG = "display";
+    auto root = nbt_new_tag_compound();
+    auto display = nbt_new_tag_compound();
+    auto name = nbt_new_tag_string(NAME.data(), NAME.size());
+    nbt_set_tag_name(name, NAME_TAG.data(), NAME_TAG.size());
+    nbt_set_tag_name(display, DISPLAY_TAG.data(), DISPLAY_TAG.size());
+    nbt_tag_compound_append(display, name);
+    nbt_tag_compound_append(root, display);
```
Trompettesib, line 86: is there an easier way to create nbt tags ? cause creating 10 variables just for one tags ... :neutral_face:
(maybe something with a json parser could be cool :upside_down_face:)
`this->_inventory->playerInventory().at(14) = protocol::Slot {true, 1, 12, nbt_tag_from_json_string("{display:{Name:'[{\"text\":\"Cubic\",\"italic\":false}]'}}")}`

STMiki: For now it will stay like this, but an nbt json parsing is required if we want to parse some commands from mc.
So yes, this will happen, but when I have the time to make a cpp wrapper for the nbt lib

STMiki:
> Btw, this happens when I try to move a block with NBT data inside my inventory, and the block disappears from my inventory too. image

I did not see that, but I think I know where I messed up 😫

cubic-server/protocol/serialization/popPrimaryType.hpp
```diff
@@ -10,9 +10,15 @@
...
 #include <cstring>
+#include <functional>
 #include <string>
 #include <vector>
 
+#include "nbt.h"
+#include "nbt.hpp"
+#include <utility>
+
+#include "nbt.hpp"
```
Trompettesib, line 21: here you included `nbt.hpp` twice

cubic-server/protocol/serialization/popPrimaryType.hpp
```diff
@@ -215,25 +221,46 @@ constexpr std::string popChat(uint8_t *&at, uint8_t *eof) { return popString(at,
...
+inline nbt_tag_t *popNbt(uint8_t *&at, uint8_t *eof)
+{
+    if (*at == (uint8_t) nbt::TagType::End) {
+        at++;
+        return nullptr;
+    }
+    auto savedAt = at;
+    auto t = nbt::parse(savedAt, eof);
+    auto diff = (savedAt - at);
+    std::cout << "diff: " << diff << std::endl;
```
Trompettesib, line 246: please remove this `std::cout`

Trompettesib (changes requested):

Trompettesib (approved):

cubic-server/protocol/serialization/popPrimaryType.hpp
```diff
@@ -215,25 +220,44 @@ constexpr std::string popChat(uint8_t *&at, uint8_t *eof) { return popString(at,
 // {
 // }
 
+static inline size_t _readMem(void *pck, uint8_t *data, size_t size)
```
huntears, line 223: A static function in a `.hpp` serves no purpose, as the function can always be accessed no matter what from any file that includes `popPrimaryType.hpp`

cubic-server/world_storage/Persistence.cpp
```diff
@@ -98,6 +98,7 @@ static size_t _readMem(void *ud, uint8_t *d, size_t s)
     _userData *u = (_userData *) ud;
     if (u->start > u->end)
         return 0;
+    // TODO(huntears) Check this +1 because this is strange uwu
```
huntears, line 101: Welp... Ok :)

huntears (changes requested):
Just one tiny thing to fix and we good

Trompettesib (approved):

huntears (approved):

PR #215, opened 2023-09-22
Fix crash when clicking outside the window

### LINKED ISSUE #193: Server crash when clicking outside of the inventory with an empty hand
When the inventory is open and the player left-click outside of the inventory box, this exception is thrown and the server crashes

image

STMiki:
which is a weird behavior of the official client

fix #193


Trompettesib (approved):
huh

STMiki:
Yep. 🙃

PR #222, opened 2023-10-09
Add a delay before picking up item

### LINKED ISSUE #217: Dropped items get picked up instantly
When dropping an item you cannot leave it on the floor as you pick it up instantly after dropping it.

huntears:
Fixes #217


huntears:
Just want to check before opening this because items drop at the wrong place and it might also happen on master.

huntears:
Welp, I found it xd

Trompettesib (commented):
there is some cringe behavior when walking around trying to get the dropped items
I suppose that we won't create grouped items for now ?

STMiki (approved):

Trompettesib (approved):

PR #226, opened 2023-10-12
Fix acacia boats droppings on every block break

### LINKED ISSUE #218: Only acacia boats drop when breaking blocks
It is currently hardcoded this way but having proper drops could be nice.

Trompettesib:
Fix #218


huntears (approved):
Need to use loot tables later on, but that should do for now indeed 💀

STMiki (approved):

PR #227, opened 2023-10-12
Fix crash caused by the entity vector being invalidated

huntears:
+ Changed the server branding in server status
+ Send all entities to players on join

### DIFF
--- a/cubic-server/Dimension.cpp
+++ b/cubic-server/Dimension.cpp
@@ -24,9 +24,11 @@ Dimension::Dimension(std::shared_ptr<World> world, world_storage::DimensionType 
 
 void Dimension::tick()
 {
-    std::lock_guard _(_entitiesMutex);
-    for (auto ent : _entities) {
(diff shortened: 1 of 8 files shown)


cubic-server/Dimension.cpp
```diff
@@ -220,24 +232,37 @@ const world_storage::ChunkColumn &Dimension::getChunk(const Position2D &pos) con
...
+                //}
+            }
+            player->sendEntityMetadata(current);
+            current.sendEntityMetadata(*player);
+        }
+    }
+    {
+        std::lock_guard _(_entitiesMutex);
+        for (auto ent : _entities) {
+            if (ent->getId() == current_id)
```
Trompettesib, line 261: Here you need to check if the entity is not a player type instead

Trompettesib (changes requested):

Trompettesib (approved):

STMiki (approved):

PR #239, opened 2023-11-10
Fix race conditions in chunk sending after gen

### LINKED ISSUE #220: Race condition on chunk generation
Sometimes an access to the chunk column map will fail and permanently lock a chunk to the void.

huntears:
This was awfully simple to fix, just a mutex to lock ;w;

Fixes #220

### DIFF
--- a/cubic-server/Dimension.cpp
+++ b/cubic-server/Dimension.cpp
@@ -170,7 +170,6 @@ void Dimension::loadOrGenerateChunk(int x, int z, const std::shared_ptr<Player> 
             return static_cast<int>(std::ceil(current_min));
         },
         [this, x, z] {
-            std::lock_guard _(_loadingChunksMutex);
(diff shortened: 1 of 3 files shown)


Trompettesib (approved):

STMiki:
Doesn't this disable multi threaded generation ?

STMiki (changes requested):
I think you could try to use a shared lock instead of a recursive one,
but as-is this completely disable multi threaded generation.
And there is [one more place](https://github.com/CubicMC/cubic-server/blob/b9a2bf444c5dd541cf725d55494ca439e7360c49/cubic-server/default/Overworld.cpp#L37) where we generate chunk

STMiki (approved):
Don't forget to open an issue

Trompettesib (approved):
looks fine

PR #242, opened 2023-11-15
Fix ore generation

NekoSelena:
Fix the generation in dirt blocks.
/!\ This fix will have to be changed when the world generation will take into account other blocks than stone as decoration.


STMiki (approved):
Don't forget to open an issue for that, because things like that are easily forgor

huntears (approved):

Issue #244, opened 2023-11-16
Item can't be pick up after being drop from inventory


STMiki:
If you drop an item by hovering the mouse over an item and hitting the drop button,
the created item entity can't be picked again

PR #245, opened 2023-11-16
Fix nbt being forgor when dropping an item

### LINKED ISSUE #236: Item Nbt tag disapear when drop
An item need to conserve it's nbt when drop, and it seems it is not the case at the moment.

Related pr: #212

STMiki:
fix #236


STMiki:
fix #244

Trompettesib (approved):
nice

huntears (approved):

PR #251, opened 2023-12-04
Add "flint and steel" item handling

NekoSelena:
> As a Player, I want to burn down my friends' houses and light the Nether portals

Definition of Done
- [x] when using the "flint and steel", a “block” of fire must be created where it is pointed when using it
- [x] durability should be updated based on usage


cubic-server/Player.cpp
```diff
@@ -1149,8 +1149,7 @@ void Player::_onUseItemOn(protocol::UseItemOn &pck)
     if (_gamemode == player_attributes::Gamemode::Creative)
         return;
     if (this->_inventory->hotbar().at(this->_heldItem).getUsabilityType() == Items::UsabilityType::RightMouseClickUsable) {
-        this->_inventory->hotbar().at(this->_heldItem).itemUse->onUse();
-        this->_inventory->hotbar().at(this->_heldItem).updateDamage();
+        this->_inventory->hotbar().at(this->_heldItem).onUse(this->getDimension(), pck.location);
         return;
```
NekoSelena, line 1153: Same as above.

cubic-server/items/usable-items/FlintAndSteel.cpp
```diff
@@ -0,0 +1,31 @@
+#include "FlintAndSteel.hpp"
+#include "Dimension.hpp"
+#include "blocks.hpp"
+
+nbt_tag_t *Items::FlintAndSteel::setNbtTag()
+{
+    int32_t DAMAGE = 60;
```
Trompettesib, line 7: this has to be set at zero, damage is the opposite of durability

STMiki (approved):

PR #260, opened 2023-12-23
Tile entities

Trompettesib:


huntears (changes requested):
Breaking a tile entity currently doesn't remove it from the list of tile entities.
You can see that by breaking the chest and placing it somewhere else then logging back in and the chest is invisible. And if you replace the chest at its original place and relogin you will see it back.

Trompettesib:
This pull request has been reworked in #263
Creating a system to handle things without those things to test the system is impossible
This is why it has been reworked when adding the furnace

Trompettesib:
This pull request has been reworked in https://github.com/CubicMC/cubic-server/pull/263
Creating a system to handle things without those things to test the system is impossible
This is why it has been reworked when adding the furnace

PR #261, opened 2023-12-27
Nether Update: Nether Wastes & Nether Portal

NekoSelena:
Known bug at the lighting of the Portal, documented here: https://github.com/CubicMC/cubic-server/issues/267


huntears (changes requested):
Most portals I tried to lit did not open up, and the only one that I got to open opened like this


The biome in the nether is a plain above y 255, which looks really weird.

I also could not open a portal in the nether normally, it would open really weirdly and then wouldn't tp me except if I was on a specific block next to the portal.

huntears:
Also I don't know if it is related but calling stop crashes the server.
image

STMiki (dismissed):

STMiki (approved):

huntears (dismissed):

cubic-server/entities/Entity.cpp
```diff
@@ -171,7 +171,17 @@ void Entity::tick()
...
-        auto block = _dim->getBlock({static_cast<int>(_pos.x), static_cast<int>(_pos.y), static_cast<int>(_pos.z)});
+
+        // here is an error arising from the '%' in the 'getBlock' below, so in negative positions, the gotten block position is 1 block less than needed
+        Position pos = _pos;
+        if (_pos.x < 0)
+            pos.x -= 1;
+        if (_pos.z < 0) {
+            pos.z -= 1;
+        }
+        auto block = _dim->getBlock(pos);
```
STMiki, line 182: then change the getBlock ?

STMiki (dismissed):

huntears (dismissed):

huntears (approved):

Trompettesib (approved):
love it !

PR #262, opened 2023-12-28
Kubik Server

Trompettesib:


huntears (changes requested):
This cannot work, the boostrap.sh script downloads to CubicServer instead of the full binary name,.
Also please have the dockerfile build the project and then use the resulting binary, which will allow for not released docker images.

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+WORKDIR /home/cubic-server
+
+COPY CMakeLists.txt .
+COPY cubic-server cubic-server
+COPY generators generators
+COPY blocks.cmake .
+COPY boost.cmake .
+COPY c_flag_overrides.cmake .
+COPY cxx_flag_overrides.cmake .
```
huntears, line 21: You already have a dockerignore, just do a `COPY . .`

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM alpine:latest AS builder
```
huntears, line 1: Please put that file at the root

docker/docker-compose.yaml
```diff
@@ -0,0 +1,8 @@
+version: "3"
```
huntears, line 1: Please put that file at the root

docker/Dockerfile
huntears: Please pin the alpine version to something more stable than latest

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY blocks.cmake .
+COPY boost.cmake .
+COPY c_flag_overrides.cmake .
+COPY cxx_flag_overrides.cmake .
+
+RUN cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DUSE_CLANG=true -DNO_GUI=true
+
+RUN cmake --build build -j`nproc --ignore=1`
+
+FROM alpine:latest AS runner
```
huntears, line 27: Please pin the alpine version to something more stable than latest

docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+# curl-dev is needed because the library is not linked statically
+RUN apk add --no-cache \
+    libstdc++ \
+    curl-dev
```
huntears, line 32: curl-dev might not be needed, curl should be enough (Not sure about that though)

huntears (changes requested):
Also doesn't build on my system
image

docker/Dockerfile
Trompettesib: It is needed
image

huntears (approved):

STMiki (approved):

PR #264, opened 2024-01-01
Fix negative coordinate blocs

ElDonad:
Getting blocks from the world storage at negative would result in incorrect ids being fetched, due to missing conversion from world position to chunk position.

A cleaner fix would be to implement a `ChunkPosition` type, which would perform the conversion automatically, and use this new type for every operation involving chunk coordinates.


Trompettesib (approved):
Finally !

huntears (approved):

PR #265, opened 2024-01-04
Arrow (Bow) handling

huntears:
This will also handle proper damages to players (Not other entities though)


Trompettesib (approved):
Sniper mode activated

NekoSelena (approved):

PR #279, opened 2024-05-26
Complete codebase rewrite

emneo-dev:
Description TBD


emneo-dev:
@Trompettesib @STMiki not sure we actually want this as a PR, we might just want this as a milestone, but I am not sure at all...

PR #297, opened 2025-04-15
Rework of the Nether Portal building algorithm

fluffyloafie:
As of today, the Nether Portal can only be built as a 4x5 full frame (with corners), and ignited from a bottom block of the inner frame.

This PR strives to optimize what can be optimized for the portal building algorithm, as well as adding the possibility to build a portal whatever its size, within the bounds set by Mojang (4x5 to 23x23)
 

It will also try to add corner-less portals support

And support all of that in negative & positive coordinates
image

EDIT 04-15-2025 ([commit](https://github.com/CubicMC/cubic-server/pull/297/commits/2ceba75ac5325a75b0a73d02d11f1ddc0cf9c27f)): At the moment, the Portal can only be ignited when a bottom obsidian block is lit. If any other block of the frame is ignited, it does nothing (except lighting a fire and... burning things ?)

EDIT 04-16-2025: At the moment, the portal can only be opened with a Flint&Steel (fire igniting the portal in any other way won't open it, only burn your feet)

EDIT 04-17-2025 ([commit](https://github.com/CubicMC/cubic-server/pull/297/commits/27f2b8fc571ec5bbe43f3b6e6768958de02a15b4)): with a partial rework of what was done during the previous days, the ignited block can be anywhere within the frame (in theory) by following this method of computation:

First computing which block within the frame was ignited

Then, after computing where the _bottomLeftCorner is, computing the direction & size of the frame, with the only step left being the filling of the portal frame if no obstruction is detected

Also, by only computing everything within the frame, the corners are never taken into account, allowing corner-less portals to be built.


fluffyloafie:
## WIP

